PR #2, opened 2018-04-18
Allow disabling IPv4

tmjd:


filesystem/etc/calico/confd/templates/bird.cfg.template
```diff
@@ -3,7 +3,8 @@ include "bird_aggr.cfg";
 include "bird_ipam.cfg";
 {{$node_ip_key := printf "/host/%s/ip_addr_v4" (getenv "NODENAME")}}{{$node_ip := getv $node_ip_key}}
 
-router id {{$node_ip}};
+{{$router_id := getenv "ROUTER_ID" ""}}
```
caseydavenport: We should probably prefix this with `CALICO_`

filesystem/etc/calico/confd/templates/bird.cfg.template
```diff
@@ -39,6 +40,7 @@ protocol direct {
   interface -"cali*", "*"; # Exclude cali* but include everything else.
 }
 
+{{if ne "" ($node_ip)}}# IPv4 disabled on this node.
```
caseydavenport: Is this comment in the right place? Won't that show up even when node_ip is defined?

pkg/startup/startup.go
```diff
@@ -421,19 +421,24 @@ func configureIPsAndSubnets(node *api.Node) (bool, error) {
...
-	} else {
+	} else if ipv6Env != "none" {
 		if ipv6Env != "" {
 			node.Spec.BGP.IPv6Address = parseIPEnvironment("IP6", ipv6Env, 6)
 		}
 		validateIP(node.Spec.BGP.IPv6Address)
 	}
 
+	if ipv4Env == "none" && (ipv6Env == "" || ipv6Env == "none") {
+		log.Warn("No IP Addresses configured or set to autodetect")
```
caseydavenport: "No IP addresses configured, and autodetection is not enabled" is probably more clear / correct.

pkg/startup/startup.go
```diff
@@ -421,19 +421,24 @@ func configureIPsAndSubnets(node *api.Node) (bool, error) {
...
+		terminate()
+	}
+
 	// Detect if we've seen the IP address change, and flag that we need to check for conflicting Nodes
-	if oldIpv4 == "" || node.Spec.BGP.IPv4Address != oldIpv4 {
+	if node.Spec.BGP.IPv4Address != oldIpv4 {
 		log.Info("Node IPv4 changed, will check for conflicts")
 		return true, nil
 	}
-	if (oldIpv6 == "" && node.Spec.BGP.IPv6Address != "") || (oldIpv6 != "" && node.Spec.BGP.IPv6Address != oldIpv6) {
```
caseydavenport: Weird... the old way just looks totally bonkers. Wonder how that happened.

tmjd: That's what I thought too. I look at it multiple times to figure out if it was handling some case that I wasn't thinking of.

PR #5, opened 2018-05-01
[WIP] [Prototype] Routereflector

caseydavenport:
Prototype of enhanced confd templates which let calico/node operate as a RR

This is a prototype of something I'm investigating now - not necessarily a thread that will come to fruition.

Goal is to provide more flexible native support for route reflector configuration via calicoctl. It does this by:
- Allowing configuration of `routeReflectorClusterID` on the Node object, which turns that node into a RR.
- Adding label selectors to BGP Peer objects to make peering configuration easier.

It relies on corresponding changes in a few places:
- libcalico - https://github.com/projectcalico/libcalico-go/pull/861
- confd - https://github.com/projectcalico/confd/pull/116

### DIFF
--- a/filesystem/etc/calico/confd/templates/bird.cfg.template
+++ b/filesystem/etc/calico/confd/templates/bird.cfg.template
@@ -61,9 +61,11 @@ template bgp bgp_template {
 
 # -------------------------------------------------
 # ------------- Node-to-node mesh -----------------
-{{- $node_cid_key := printf "/host/%s/cluster_id" (getenv "NODENAME")}}{{if exists $node_cid_key}}
+{{- $node_cid_key := printf "/host/%s/cluster_id" (getenv "NODENAME")}}
+{{- $node_cluster_id := getv $node_cid_key}}
+{{- if ne "" ($node_cluster_id)}}
 
-# This is a route reflector node.
+# This node is configured as a RR in $node_cluster_id. Ignore node-to-node mesh setting.
 {{- else}}
 {{- if (json (getv "/global/node_mesh")).enabled}}
 {{- range $host := lsdir "/host"}}
@@ -74,9 +76,11 @@ template bgp bgp_template {
 # For peer {{$onode_ip_key}}
 {{- if eq $onode_ip ($node_ip) }}
 # Skipping ourselves ({{$node_ip}})
-{{else if ne "" $onode_ip}}protocol bgp Mesh_{{$id}} from bgp_template {
+{{- else if ne "" $onode_ip}}
+protocol bgp Mesh_{{$id}} from bgp_template {
   neighbor {{$onode_ip}} as {{if exists $onode_as_key}}{{getv $onode_as_key}}{{else}}{{getv "/global/as_num"}}{{end}};
-}{{- end}}{{end}}{{end}}
+}
+{{- end}}{{end}}{{end}}
 {{- else}}
 # Node-to-node mesh disabled
 {{- end}}


filesystem/etc/calico/confd/templates/bird.cfg.template
```diff
@@ -59,43 +59,116 @@ template bgp bgp_template {
...
 
-# ------------- Node-to-node mesh -------------
-{{if (json (getv "/global/node_mesh")).enabled}}
-{{range $host := lsdir "/host"}}
-{{$onode_as_key := printf "/host/%s/as_num" .}}
-{{$onode_ip_key := printf "/host/%s/ip_addr_v4" .}}{{if exists $onode_ip_key}}{{$onode_ip := getv $onode_ip_key}}
-{{$nums := split $onode_ip "."}}{{$id := join $nums "_"}}
+# -------------------------------------------------
+# ------------- Node-to-node mesh -----------------
+{{- $node_cid_key := printf "/host/%s/cluster_id" (getenv "NODENAME")}}{{if exists $node_cid_key}}
```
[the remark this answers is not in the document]
caseydavenport: Yes, I agree... I've tried to use indentation to make it a bit more clear (rather than just a big block). 

I'll see if there's any other formatting magic (or comments) I can add to the code to make it more clear what's going on.

caseydavenport:
Pushed a build of this to `caseydavenport/node:rr2`.

You can use try it out with something like this:

- Pull down the latest `master` manifest, replacing the image with the one above and restarting pods.
- Build this calicoctl PR so you can use it in the steps below: https://github.com/projectcalico/calicoctl/pull/1882
- `calicoctl node status` should show node-node mesh working.
- Disable node-to-node-mesh, `calicoctl node status` should show no peers.
- Add a simple selector-based peer, like the one below which just replicates node-node mesh behavior.

```
apiVersion: projectcalico.org/v3
kind: BGPPeer
metadata:
  name: node-mesh 
spec:
  nodeSelector: "all()"
  peerSelector: "all()"
```

- You should now be able to see all the nodes established again.

^ The above is just a basic test of the label selection functionality. 

You can then make one of the nodes a route reflector by setting `routeReflectorClusterID: 10.0.0.1` on one of the nodes. BGP status should remain established.

If you view its BIRD config, you should see the right RR config present:

```
# -------------------------------------------------
# ------------ Configured peers -------------------

# BGPPeer node-mesh
# map[peer_selector:all() name:node-mesh ip: as_num:0 node_selector:all()]

# Calico node casey-crc-kadm-ms
# Skipping ourselves (10.128.0.4)

# Calico node casey-crc-kadm-node-0
protocol bgp Node_10_128_0_5 from bgp_template {
  # We're a RR but our peer isn't - act as a client.
  neighbor 10.128.0.5 as 64512;
  rr client;
  rr cluster id 10.0.0.1;
}

# Calico node casey-crc-kadm-node-1
protocol bgp Node_10_128_0_6 from bgp_template {
  # We're a RR but our peer isn't - act as a client.
  neighbor 10.128.0.6 as 64512;
  rr client;
  rr cluster id 10.0.0.1;
}
```

filesystem/etc/calico/confd/templates/bird.cfg.template
```diff
@@ -59,43 +59,120 @@ template bgp bgp_template {
...
 
-# ------------- Global peers -------------
-{{if ls "/global/peer_v4"}}
-{{range gets "/global/peer_v4/*"}}{{$data := json .Value}}
-{{$nums := split $data.ip "."}}{{$id := join $nums "_"}}
-# For peer {{.Key}}
-protocol bgp Global_{{$id}} from bgp_template {
+# Peer with IP {{$data.ip}}
+      {{- $nums := split $data.ip "."}}{{$id := join $nums "_"}}
+      {{- $node_cid_key := printf "/host/%s/cluster_id" (getenv "NODENAME")}}{{if exists $node_cid_key}}{{$cluster_id := getv $node_cid_key}}
```
caseydavenport: There are a bunch of places in this config where we re-query the cluster_id - I don't think this is needed any more, since we store it off in a variable called `node_cluster_id`. Might be able to clean that up a bit.

filesystem/etc/calico/confd/templates/bird.cfg.template
```diff
@@ -59,43 +59,120 @@ template bgp bgp_template {
...
+  rr cluster id {{$cluster_id}};
 }
-{{end}}
-{{else}}# No node-specific peers configured.{{end}}
-{{end}}{{/* End of IPv4 enable check */}}
+                  {{- else}}
+
+# Calico node {{.}}
+protocol bgp RR_{{$id}} from bgp_template {
+  # We're both RRs, don't act as a client.
```
caseydavenport: I think we may need to check if the RR is within our cluster or not - if it's not in our cluster, then we probably need to peer with it as a client. 

Need to double check this though.

neiljerram:
Note: being replaced by #50

PR #11, opened 2018-05-31
Don't specify an explicit source address

caseydavenport:
## Description

Need to explore what the impacts are, but might make configuring Calico in multi-interface scenarios easier. 

CC @kprabhak 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


caseydavenport:
My concern is that we might break users who are relying on the IP_AUTODETECT functionality to pick a specific source address (or who have hard-coded a specific addr in the node object).

I think we might need a way of indicating "don't use a set source addr" in the node object. Perhaps with another field, annotation, or special value.

Issue #9, opened 2018-05-25
If allocate-ipip-addr fails (due to expected race), it doesn't get retried


ti-mo:
Mentioned in https://github.com/projectcalico/libcalico-go/pull/785.

## Expected Behavior

Declaring an IP pool on an existing deployment should allocate a block to every cluster member.

## Current Behavior

```
2018-04-27T08:49:12Z public-slave-14 confd[130]: ERROR "
2018-04-27 08:49:11.878 [INFO][277] client.go 202: Loading config from environment
2018-04-27 08:49:11.895 [INFO][277] ipam.go 120: Auto-assign 1 ipv4, 0 ipv6 addrs for host 'public-slave-14'
2018-04-27 08:49:11.903 [INFO][277] ipam.go 172: Ran out of existing affine blocks for host 'public-slave-14'
2018-04-27 08:49:11.909 [INFO][277] ipam.go 195: Need to allocate 1 more addresses - allocate another block
2018-04-27 08:49:11.932 [INFO][277] ipam_block_reader_writer.go 116: Claiming a new affine block for host 'public-slave-14'
2018-04-27 08:49:11.944 [INFO][277] ipam_block_reader_writer.go 159: Host public-slave-14 claiming block affinity for 10.123.84.64/26
2018-04-27 08:49:11.994 [WARNING][277] ipam_block_reader_writer.go 187: Problem claiming block affinity:%!!(MISSING)(EXTRA errors.ErrorResourceAlreadyExists=resource already exists: BlockKey(cidr=10.123.84.64/26))
2018-04-27 08:49:12.060 [ERROR][277] ipam.go 204: Error claiming new block: 10.123.84.64/26 already claimed by %!!(MISSING)s(*string=0xc420283430)
2018-04-27 08:49:12.060 [ERROR][277] ipam.go 135: Error assigning IPV4 addresses: 10.123.84.64/26 already claimed by %!!(MISSING)s(*string=0xc420283430)
2018-04-27 08:49:12.060 [FATAL][277] allocate_ipip_addr.go 120: Unable to autoassign an address for IPIP error=10.123.84.64/26 already claimed by %!!(MISSING)s(*string=0xc420283430)
"
2018-04-27T08:49:12Z public-slave-14 confd[130]: ERROR exit status 1
```

[...]

PR #18, opened 2018-06-12
allocate-ipip-addr: retry loop around IPAM AutoAssign()

ti-mo:
Has a random sub-second delay to prevent the (unlikely) perpetuation
of the race condition throughout the retry loop.

## Description

Workaround for the race condition outlined in https://github.com/projectcalico/node/issues/9. All relevant details are mentioned there.

I don't have sufficient knowledge of the project to write a meaningful unit test. The 'why' is explained in the code.

## Todos
- [ ] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Internally retry address allocation from IP Pools with IPIP enabled (impacts pre-3.0 releases)
```

@caseydavenport

### DIFF
--- a/pkg/allocateipip/allocate_ipip_addr.go
+++ b/pkg/allocateipip/allocate_ipip_addr.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"math/rand"
 	"os"
 	"time"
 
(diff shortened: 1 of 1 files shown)


ti-mo:
@caseydavenport Since this is likely only useful for 2.6, should it be backported to https://github.com/projectcalico/calico `release-v2.6`?

pkg/allocateipip/allocate_ipip_addr.go
```diff
@@ -159,9 +160,20 @@ func assignHostTunnelAddr(ctx context.Context, c client.Interface, nodename stri
...
+	// Retry loop around AutoAssign to recover from a distributed race in case
+	// all nodes in a large cluster respond to a change in /pools simultaneously.
+	for a := 3; a > 0; a-- {
+		var err error
+		ipv4Addrs, _, err = c.IPAM().AutoAssign(ctx, args)
+
+		if err != nil {
+			log.WithError(err).Error("Unable to autoassign an address for IPIP:")
+			time.Sleep(time.Duration(rand.Intn(1000)) * time.Millisecond)
+		} else {
```
caseydavenport: minor nit: I don't think we need the else block around the break here, so I'd be in favor of removing it to make reading easier.

caseydavenport:
> Since this is likely only useful for 2.6 or earlier, should it be backported to https://github.com/projectcalico/calico release-v2.6?

@ti-mo yep, I think it should be. Unfortunately the `calico/node` code moved repos between v3.1 and (upcoming) v3.2, so you'll need to open a PR over in the Calico repo.

As for this PR, I'm not sure it's needed? I believe that in v3 the IPAM code does the necessary retries.

caseydavenport:
Example retry in v3: https://github.com/projectcalico/libcalico-go/blob/master/lib/ipam/ipam.go#L281-L285

In v2.6, we don't retry: https://github.com/projectcalico/libcalico-go/blob/release-v2.6/lib/client/ipam.go#L190-L223

pkg/allocateipip/allocate_ipip_addr.go
ti-mo: But of course, will fix in the new PR.

ti-mo:
Closing in favor of https://github.com/projectcalico/calico/pull/2038.

PR #20, opened 2018-06-18
Stop running etcd SSL tests

tomdee:
They are tested by K8s e2es instead


caseydavenport:
@tomdee approved. I'd still like to trim some more from these tests in order to get the build time down sub 20 minutes.

If you've got any ideas, do let me know :)

PR #21, opened 2018-06-21
[DO NOT MERGE] Add some super simple timings to the Makefile

tomdee:


tomdee:
Some noes:
```
Calico/node timings                                                              
65s - creating vendor (and pulling go-build)                                          
2s to download etcd                                                                   
10s to download hyperkube                                                             
52s FV (should be much quicker)                                                       
30ish s - building binaries                                                           
a few seconds each - pulling binaries (felix etc...)                                       
22s - bulding calico/node                                                             
6s - building the workload                                                            
18s - building calico/test                                                            
1920s - running the STs       
```

I had thought it would be good to not build calico/test every time (since it's not changed in >1year), but might not be worth it to save 18s
Running the FVs takes a lot longer than it should - there's probably some simple Go caching we could do to save >30s

But the biggest thing is the STs.

PR #24, opened 2018-06-25
Reinstate an SSL test

neiljerram:


caseydavenport:
Going to test this via a k8s e2e so we can get manifest coverage included, which is where the majority of the risk is.

PR #25, opened 2018-07-02
Update release process

caseydavenport:
## Description

this PR aligns the release targets / RELEASING.md with other repos. It also added in a couple of other things to make this easier:
- Rename the produced image to `calico/node:latest-$(ARCH)` to match elsewhere.
- Add in CD target.
- Add gcr.io image building for GKE. 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,7 +94,7 @@ clean:
 	rm -rf certs *.tar vendor $(NODE_CONTAINER_BIN_DIR)
 
 	# Delete images that we built in this repo
-	docker rmi $(NODE_CONTAINER_NAME):latest || true
+	docker rmi $(NODE_CONTAINER_NAME):latest-$(ARCH) || true
 	docker rmi $(TEST_CONTAINER_NAME) || true
 
 	# Retag and remove external images so that they will be pulled again
@@ -163,7 +163,7 @@ $(NODE_CONTAINER_CREATED): ./Dockerfile$(ARCHTAG) $(NODE_CONTAINER_FILES) $(addp
 	  echo; echo calico-bgp-daemon -v;   /go/bin/calico-bgp-daemon -v; \
 	  echo; echo confd --version;        /go/bin/confd --version; \
 	"
-	docker build --pull -t $(NODE_CONTAINER_NAME) . --build-arg ver=$(CALICO_GIT_VER) -f ./Dockerfile$(ARCHTAG)
+	docker build --pull -t $(NODE_CONTAINER_NAME):latest-$(ARCH) . --build-arg ver=$(CALICO_GIT_VER) -f ./Dockerfile$(ARCHTAG)
 	touch $@
 
 # Get felix binaries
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -213,6 +213,59 @@ $(NODE_CONTAINER_BIN_DIR)/birdcl:
...
+imagetag:
+ifndef IMAGETAG
+	$(error IMAGETAG is undefined - run using make <target> IMAGETAG=X.Y.Z)
+endif
+
+## push one arch
+push: imagetag
+	docker push $(NODE_CONTAINER_NAME):$(IMAGETAG)-$(ARCH)
+	docker push quay.io/$(NODE_CONTAINER_NAME):$(IMAGETAG)-$(ARCH)
+ifeq ($(ARCH),amd64)
```
fasaxc: Looks like you tag the -ARCH versions for gcr but don't push them

Makefile
```diff
@@ -213,6 +213,59 @@ $(NODE_CONTAINER_BIN_DIR)/birdcl:
...
+endif
+endif
+
+push-all: imagetag $(addprefix sub-push-,$(ARCHES))
+sub-push-%:
+	$(MAKE) push ARCH=$* IMAGETAG=$(IMAGETAG)
+
+## tag images of one arch
+tag-images: imagetag
+	docker tag $(NODE_CONTAINER_NAME):latest-$(ARCH) $(NODE_CONTAINER_NAME):$(IMAGETAG)-$(ARCH)
```
fasaxc: Feels like this might be getting out of control.  Could we have a variable that contains the list of repos to push to:
```
DOCKER_REMOTES:=quay.io/ 
ifeq ($(RELEASE),true)
DOCKER_REMOTES+=gcr.io/ eu.gcr.io/ ...
endif
```
and then loop over it in `tag-images` and `push`?

Makefile
```diff
@@ -388,64 +441,120 @@ st: dist/calicoctl busybox.tar routereflector.tar calico-node.tar workload.tar r
 	           -e HOST_CHECKOUT_DIR=$(CURDIR) \
 	           -e DEBUG_FAILURES=$(DEBUG_FAILURES) \
 	           -e MY_IP=$(LOCAL_IP_ENV) \
-	           -e NODE_CONTAINER_NAME=$(NODE_CONTAINER_NAME) \
-             -e RR_CONTAINER_NAME=$(RR_CONTAINER_NAME):$(RR_VER) \
+		   -e NODE_CONTAINER_NAME=$(NODE_CONTAINER_NAME):latest-$(ARCH) \
```
fasaxc: Change to the indent here (2 tabs) makes it inconsistent with the line above (1 tab then spaces). The latter is my preferred style since it lets you align arguments visually.

Makefile
```diff
@@ -388,64 +441,120 @@ st: dist/calicoctl busybox.tar routereflector.tar calico-node.tar workload.tar r
...
+                   -e RR_CONTAINER_NAME=$(RR_CONTAINER_NAME):$(RR_VER) \
```
fasaxc: This line is different again (all spaces)

Makefile
caseydavenport: SGTM - one thing that makes this a bit trickier is that the org is different across different repos.

e.g. calico vs projectcalico-org.

I've done that by putting that in the list as well, but it means the NODE_CONTAINER_NAME override won't work for the re-tagged images. I think that is OK - but LMK if there's a reason that doesn't work.

caseydavenport:
@fasaxc thanks for the feedback - I think it's ready for another look :)

caseydavenport:
If this looks good I'll port the tag/push changes to typha and CNI

caseydavenport:
@fasaxc WDYT about these latest changes?

PR #26, opened 2018-07-05
Build a single calico-node binary by importing code

caseydavenport:
## Description

Something I wanted to try - this modifies the calico/node build to produce a single `calico-node` binary, removing the need to produce the following separate binaries:
- calico-felix
- startup
- allocate-ipip-addr
- check-readiness
- confd (needs this PR: https://github.com/projectcalico/confd/pull/132)

Instead, you get something like this:

```
./bin/calico-node -h
Usage of Calico:
  -allocate-ipip-addr
        Allocate an IPIP address for this node
  -bird-ready
        Run BIRD readiness checks
  -bird6-ready
        Run BIRD6 readiness checks
  -confd
        Run confd
  -confd-keep-stage-file
        Keep stage file when running confd
  -confd-run-once
        Run confd in oneshot mode
  -felix
        Run Felix
  -felix-ready
        Run felix readiness checks
  -startup
        Initialize a new node
  -v    Display version
flag: help requested
```

This has some benefits:
- Reduces makefile and build complexity a bunch.
- Reduces image size. Without this PR calico/node is 248 MB. With it, it's 70MB.
- We're now pinning dependencies like Felix and confd, so it gives an easier and more consistent way to trace dependency versions.

I've removed `calico-bgp-daemon` from the build for now, since it doesn't work in v3.X anyway. Once it does, we can add it back in as a vendor dependency rather than a separate binary. 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Substantially reduced the calico/node image size by consolidating shared code in a single binary.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,18 +31,10 @@ CALICO_BUILD?=calico/go-build$(ARCHTAG):$(GO_BUILD_VER)
 CALICO_GIT_VER := $(shell git describe --tags --dirty --always)
 
 # Versions and location of dependencies used in the build.
-CONFD_VER?=master
-FELIX_VER?=master
 BIRD_VER?=v0.3.2
-GOBGPD_VER?=v0.2.1
-FELIX_REPO?=calico/felix
-FELIX_CONTAINER_NAME?=$(FELIX_REPO)$(ARCHTAG):$(FELIX_VER)
-CONFD_REPO?=calico/confd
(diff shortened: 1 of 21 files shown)


caseydavenport:
Tests are failing because `calicoctl node status` expects to find a `calico-felix` process running. Need to think through how to fix that.

PR #34, opened 2018-07-29
Standard Dockerfiles, build and push for all archs

deitch:
## Description

Several pieces to this PR:

* build and distribute binaries and images for all supported arches
* Remove `for` loop that would cause silent failure if one deploy succeeded but others didn't (`for` loops in sh in Makefile are *not* `set -x` by default)
* build binaries and build and deploy images for all supported arches by default as part of `make cd`, with `EXCLUDEARCH` to exclude any with issues (as of now, `s390x`)
* Build binaries for `calico-node` in `dist/bin/` rather than `filesystem/bin/` so that it is easier to build images
* Use `calico/bird` image in a multistage `Dockerfile` for each arch rather than trying to `curl` it. Makes it simpler, but also picks up correct arch
* Standardize `Dockerfile.<arch>` names per all of the other calico projects
* Parametrize the qemu image (from `calico/go-build`) and bird image (from `calico/bird`) in each `Dockerfile`
* Add `.calico-node.created-<arch>` for every arch to `.gitignore`

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

cc @tomdee @fasaxc


tomdee:
Overall, this looks good but I'll need to review in more detail later in the week. One area that I want to think about some more is the change to start the downloading of the bird binaries the Dockerfiles instead of in the Makefile.

deitch:
> change to start the downloading of the bird binaries the Dockerfiles instead of in the Makefile

Yeah, @fasaxc and I had a conversation about this on Slack when we started it. There were a few reasons that led us to lean in this direction:

* Only `amd64` binaries on GitHub. Granted, this can be fixed, but it wasn't available now. 
* More importantly (related to the previous), the ease of getting binaries distributed to docker hub in an automated fashion vs to GitHub releases. 
* On docker hub, we also include interim builds as different tags, which makes it easy to pick exactly which binaries we wanted.
* Natural flow with Dockerfiles when building images. Since we are downloading bird to use in another image, use the built-in way of doing it rather than the interim download step.

fasaxc:
On the face of it, (as long as we have that build arg to set the bird version) using the bird image as a `FROM xxx` seems neat to me; might help make sure that bird is always "fresh"  (whereas combining make and curl results in stale builds since there's no way to give a sensible make dependency).  I.e.

```
filesystem/bin/bird: <can't put anything sensible here>
    curl $(BIRD_URL)
```

Makefile
```diff
@@ -166,60 +211,58 @@ $(NODE_CONTAINER_BIN_DIR)/calico-node: vendor
 ###############################################################################
 # Building the image
 ###############################################################################
-## Create the calico/node image.
-image: $(NODE_CONTAINER_NAME)
-$(NODE_CONTAINER_NAME): $(NODE_CONTAINER_CREATED)
-$(NODE_CONTAINER_CREATED): ./Dockerfile$(ARCHTAG) $(NODE_CONTAINER_FILES) $(addprefix $(NODE_CONTAINER_BIN_DIR)/,$(NODE_CONTAINER_BINARIES))
+## Create the calico/node image for the current ARCH
+image: $(CONTAINER_NAME)
+## Create the calico/node imagesfor all supported ARCHes
```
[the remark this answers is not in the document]
deitch: Fixed.

Makefile
```diff
@@ -7,63 +7,108 @@ all: build
...
+ifeq ($(ARCH),aarch64)
+        override ARCH=arm64
+endif
+ifeq ($(ARCH),x86_64)
+    override ARCH=amd64
 endif
 
-ifeq ($(ARCH),s390x)
-	ARCHTAG:=-s390x
+# so we can use filenames in targets
```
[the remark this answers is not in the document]
deitch: Sure, happy to add it. Coming shortly...

deitch:
I updated the comments based on your requests @fasaxc 

As for your earlier comment:

> I tried it locally and hit an issue with the image-all target not working

The way that it works is:

1. `image-all` target has a dependency on all of `sub-image-<each_arch>`
2. Each `sub-image-%` target just has a recipe for `make image ARCH=$*`
3. `image` target has a dependency on `$(CONTAINER_NAME)`, which is just `calico/node`
4. `calico/node` has a dependency on `$(NODE_CONTAINER_CREATED)`, which is just the flag file `.calico_node.created-$(ARCH)`

So, can you check if the flag file exists, and try to `make -n` for each target there?

Makefile
```diff
@@ -7,63 +7,112 @@ all: build
...
-# # When ARCH is undefined it defaults to amd64.
-ARCH?=amd64
-ifeq ($(ARCH),amd64)
-	ARCHTAG?=
+###############################################################################
+# Both native and cross architecture builds are supported.
+# The target architecture is select by setting the ARCH variable.
+# When ARCH is undefined it is set to the detected host architecture.
+# When ARCH differs from the host architecture a crossbuild will be performed.
+ARCHES=$(patsubst docker-image/Dockerfile.%,%,$(wildcard docker-image/Dockerfile.*))
```
tomdee: This needs the `docker-image/` part to be removed

Makefile
```diff
@@ -7,63 +7,112 @@ all: build
...
+
 K8S_VERSION?=v1.10.4
 HYPERKUBE_IMAGE?=gcr.io/google_containers/hyperkube-$(ARCH):$(K8S_VERSION)
 TEST_CONTAINER_FILES=$(shell find tests/ -type f ! -name '*.created')
 
+# Variables controlling the calico/node image
+NODE_CONTAINER_CREATED=.calico_node.created-$(ARCH)
+NODE_CONTAINER_BIN_DIR=./dist/bin/
+# files in the container that we create locally
+NODE_BINARIES_FILES=calico-node-$(ARCH)
```
tomdee: This says FILES but it's only ever one file.

Makefile
```diff
@@ -7,63 +7,112 @@ all: build
...
+NODE_CONTAINER_BINARIES = $(addprefix $(NODE_CONTAINER_BIN_DIR),$(NODE_BINARIES_FILES))
```
tomdee: And this is only ever one file, so shouldn't be plural. I think You can just get rid of NODE_BINARIES_FILES and just put the value inline here

deitch:
@tomdee thanks for fixing it up while I was on vacation (holiday for our British friends).

PR #36, opened 2018-08-08
Makefile: Ensure register is called before cross-building

tomdee:


tomdee:
Merging even though CI isn't passing since this PR causes CI to get further than before, and the failure hit here could just be a flake.

PR #44, opened 2018-08-22
Remove need for NODE_VERSIONS

caseydavenport:
## Description

Don't need this any more since `calico-node` can report its own version. 

Also do some .gitignore cleaning. 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


.gitignore
```diff
@@ -1,24 +1,12 @@
...
-venv
-env
+dist/*
 certs
 .idea
 *.pyc
-*~
 .coverage
-cover
-*.tar
```
tomdee: Are you sure about removing .tar?

.gitignore
```diff
@@ -1,24 +1,12 @@
...
 *.created-*
-docker
 **/*.sw[pon]
-birdcl
-.vagrant
-gobgp
-calicoctl/calicoctl
-bin
```
tomdee: And won't about bin and release? We still put files in those directories that we don't want checked in right?

caseydavenport: I don't think we put anything in either of those directories - looks like the binaries are built in `dist` and the release directory hasn't been used in a long time, AFAICT.

.gitignore
caseydavenport: Nope, we should keep that.

PR #50, opened 2018-09-04
Template and testing updates for new RR design

neiljerram:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,10 +99,8 @@ BIRD_IMAGE ?= calico/bird:$(BIRD_VER)-$(ARCH)
 # Versions and locations of dependencies used in tests.
 CALICOCTL_VER?=master
 CNI_VER?=master
-RR_VER?=master
 TEST_CONTAINER_NAME_VER?=latest
 CTL_CONTAINER_NAME?=calico/ctl:$(CALICOCTL_VER)-$(ARCH)
-RR_CONTAINER_NAME ?= calico/routereflector:$(RR_VER)
 TEST_CONTAINER_NAME?=calico/test:$(TEST_CONTAINER_NAME_VER)-$(ARCH)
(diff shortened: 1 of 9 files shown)


neiljerram:
@caseydavenport There are pinning conflicts and updates that will be needed before this can be merged, but that's all mechanical stuff, and I'd appreciate your review of the substantive changes here.

neiljerram:
Thanks @caseydavenport for your review.  As noted in the confd PR, I think the key factor for your suggestion is whether we want to enhance the calicoctl UI now, and if so how - so please let's talk about that.

caseydavenport:
@neiljerram I think, as you suggested, we should keep as-is for now. We can make incremental improvements to the calicoctl UI, etc as follow on work if so desired. I'd rather not hold this up on that!

Makefile
```diff
@@ -175,7 +173,7 @@ LIBCALICO_VERSION?=$(shell git ls-remote git@github.com:projectcalico/libcalico-
 FELIX_REPO?=github.com/projectcalico/felix
 FELIX_VERSION?=$(shell git ls-remote git@github.com:projectcalico/felix master 2>/dev/null | cut -f 1)
 CONFD_REPO?=github.com/projectcalico/confd
-CONFD_VERSION?=$(shell git ls-remote git@github.com:projectcalico/confd master 2>/dev/null | cut -f 1)
+CONFD_VERSION?=$(shell git ls-remote git@github.com:projectcalico/confd feature-internal-rr 2>/dev/null | cut -f 1)
```
caseydavenport: Need to update this to master once merged.

caseydavenport: I meant to say once the confd PR was merged, which is has now, so we can update this :)

PR #55, opened 2018-09-13
Streamline templating for 'rr client' vs 'next hop self'

neiljerram:
```release-note
calico/node no longer configures next hop self on all BGP peer configurations. 
```

### DIFF
--- a/filesystem/etc/calico/confd/templates/bird.cfg.template
+++ b/filesystem/etc/calico/confd/templates/bird.cfg.template
@@ -92,14 +92,9 @@ template bgp bgp_template {
 {{- else}}
 protocol bgp Global_{{$id}} from bgp_template {
   neighbor {{$data.ip}} as {{$data.as_num}};
-{{- if ne "" ($node_cluster_id)}}
-{{- if ne $data.rr_cluster_id ($node_cluster_id)}}
+{{- if and (ne "" ($node_cluster_id)) (ne $data.rr_cluster_id ($node_cluster_id))}}
   rr client;
   rr cluster id {{$node_cluster_id}};
-{{- else}}
-  next hop self;     # Disable next hop processing and always advertise our
-                     # local address as nexthop
-{{- end}}
 {{- else}}
   next hop self;     # Disable next hop processing and always advertise our
                      # local address as nexthop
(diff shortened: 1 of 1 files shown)


redbaron:
It doesn't seem to have an effect, I have to add "next hop keep" to template to make it stop advertising self over eBGP, seems that default value is not what bird documentation says (or could it be your custom patch?)

neiljerram:
@redbaron I'm afraid there's a bit of detail here that I don't understand.  (I'm not yet an expert in reading birdc output.)

Your comment is about eBGP behaviour, but then you're focusing on advertisement between two nodes in the same cluster via full-mesh peering, which is necessarily iBGP.  Have I misunderstood something?

I wonder if this is not about eBGP, but about a node having multiple IPs.  Are the IPs 172.24.5.211 and 172.24.5.102 on the same node?  If not, how are they related to the node hosting that route (172.22.8.192/26) and the node that you're observing from?

By the way, on an eBGP peering, `next hop self` behaviour is typically required, isn't it?  (Because you wouldn't expect all of the internal IPs of one AS to be resolvable from a different AS.)

redbaron:
> Your comment is about eBGP behaviour, but then you're focusing on advertisement between two nodes in the same cluster via full-mesh peering, which is necessarily iBGP. Have I misunderstood something?


```
bird> show route for 172.22.8.192/26 all export Node_172_25_5_212
172.22.8.192/26    via 172.24.5.211 on ens192 [Mesh_172_24_5_211 11:21:35]
```

is `172.22.8.192/26` route learned from internal `Mesh_172_24_5_211` mesh peering, but it is being advertised to `Node_172_25_5_212` peer which is a node in another cluster with a different AS number (so it is eBGP if I understand it).

> By the way, on an eBGP peering, next hop self behaviour is typically required, isn't it? (Because you wouldn't expect all of the internal IPs of one AS to be resolvable from a different AS.)

Maybe, I am not a networking guy and might miss something. I am looking at the case where subset of nodes are peering with each other and exchange full set of routes they know, all of it while every node can speak to every node (the reason all nodes are not peering with each other is purely because majority of them have dynamic IP). It is very much likely as you described that default behaviour for  eBGP is `next hop self` as it is indeed safer

redbaron:
> OK, so I think the situation is like this:

your diagram is accurate.

> Isn't it unusual, though, for nodes in different ASs to be directly reachable, if they aren't eBGP peers themselves?

in the  networking world it might be unusual, but in cloud/kubernetes case reason they are not eBGP peers has nothing to do with their reachibility :)

Anyway, I can see that `next hop` value seems to default to `self` in  eBGP and there is a good  historical reason to do it. I all update https://github.com/projectcalico/calico/issues/1942#issuecomment-451745429 to mention it

PR #75, opened 2018-09-28
Add option to build from local felix/confd/libcalico

caseydavenport:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


caseydavenport:
Once we're happy with this approach I think we can roll this out to other repos as well (though most won't need felix+confd, just libcalico).

PR #87, opened 2018-10-24
Factor out OS-specific function from startup.go.

fasaxc:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/startup/startup.go
+++ b/pkg/startup/startup.go
@@ -55,14 +55,6 @@ const (
 	AUTODETECTION_METHOD_SKIP_INTERFACE = "skip-interface="
 )
 
-// Default interfaces to exclude for any logic following the first-found
-// auto detect IP method
-var DEFAULT_INTERFACES_TO_EXCLUDE []string = []string{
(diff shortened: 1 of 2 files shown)


pkg/startup/startup.go
```diff
@@ -282,17 +274,26 @@ func determineNodeName() string {
 	return nodeName
 }
 
-// nodenameFromFile reads the /var/lib/calico/nodename file if it exists and
+func nodenameFileName() string {
+	fn := os.Getenv("CALICO_NODENAME_FILE")
```
caseydavenport: Do we really want this to be configurable?

What about we have a `nodenameFile` defined in `startup_linux.go` which simply returns `/var/lib/calico/nodename` always?

I think setting CALICO_NODENAME_FILE to something else right now will break the CNI plugin anyway.

fasaxc:
Agreed, but I think this will be needed on other platforms so I think we agreed to live with...

PR #90, opened 2018-10-26
Disown confd config

neiljerram:
Together with https://github.com/projectcalico/confd/pull/165, this PR moves Calico's confd config and templates from this (node) repo to the confd repo.  I think this will be a big win for confd template work, as currently we have to modify code, templates and expected config in two mutually-dependent repos.

With this change, there is still one piece of annoying reverse dependence: `etc/calico/confd/conf.d/tunl-ip.toml` (now in confd) includes `reload_cmd = "calico-node -allocate-ipip-addr"`.  But that doesn't affect any CI, and I believe we can live with it.


neiljerram:
@caseydavenport @briansan I'm sorry, I didn't mean to merge this with that "For PR testing..." commit still present; and as a result node CI has broken.  Just fixing that with a follow-up PR now...

neiljerram:
... #92 should be the fix for that hastiness.

PR #94, opened 2018-11-04
Automatic update of libcalico-go to 9d3859e1544309049df51602b484031ee201e31a

doublek:


doublek:
Sigh. Looks like gometalinter is checking packages in the vendor directory and not ignoring it like we ask it to. Looking into it.

PR #95, opened 2018-11-04
Automatic update of libcalico-go to 489a4d4fcdf275caca1edf59e95054b6e8601555

doublek:


doublek:
Sigh. Looks like gometalinter is checking packages in the vendor directory and not ignoring it like we ask it to. Looking into it.

doublek:
glide was pulling in the an outdated semver package. Re running glide forced ended up picking the version that it should've in the first place.

PR #97, opened 2018-11-06
Add new kubeadm-dind-cluster ST and basic service advertisement test

lwr20:
Prototype, do not merge


neiljerram:
@lwr20 Can we have a clearer PR title?  Something like "Add new kubeadm-dind-cluster ST"

Makefile
```diff
@@ -451,6 +451,43 @@ st-checks:
 	# running on the host.
 	iptables-save | grep -q 'calico-st-allow-etcd' || iptables $(IPT_ALLOW_ETCD)
 
+.PHONY: k8s-test
+## Run the k8s tests
+k8s-test: $(NODE_CONTAINER_CREATED) calico_test.created
+	wget https://raw.githubusercontent.com/lwr20/kubeadm-dind-cluster/master/fixed/dind-cluster-v1.10.sh
```
lwr20: Not a good plan to rely on this repo...

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -0,0 +1,217 @@
...
+        _log.debug("Service created. status='%s'" % str(api_response.status))
+
+        # # Modify calico/node to run :master tag and set CALICO_STATIC_ROUTES=10.96.0.0/12
+
+        config.load_kube_config(os.environ.get('KUBECONFIG'))
+        api = client.AppsV1Api(client.ApiClient())
+        node_ds = api.read_namespaced_daemon_set("calico-node", "kube-system", exact=True, export=True)
+        for container in node_ds.spec.template.spec.containers:
+            if container.name == "calico-node":
+                container.image = "calico/node:latest-amd64"
```
lwr20: Feels like changing the calico/node image to be the one just built should be a thing that happens for ALL tests in here, not just this one.  

Consider moving it to cluster setup steps.

calico_test/Dockerfile.amd64.calico_test
```diff
@@ -51,5 +51,7 @@ RUN wget https://github.com/coreos/etcd/releases/download/v2.3.3/etcd-v2.3.3-lin
     cd etcd-v2.3.3-linux-amd64 && \
     ln -s etcdctl /usr/local/bin/
 
+COPY * /code/
```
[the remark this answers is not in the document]
lwr20: The python kubernetes client package needs these.

Makefile
```diff
@@ -451,6 +451,43 @@ st-checks:
...
+k8s-start:
+	CNI_PLUGIN=calico ./dind-cluster-v1.10.sh up
+	docker save $(BUILD_IMAGE):latest-$(ARCH) > calico-node.tar
+	docker cp calico-node.tar kube-master:/calico-node.tar
+	docker cp calico-node.tar kube-node-1:/calico-node.tar
+	docker cp calico-node.tar kube-node-2:/calico-node.tar
+	docker exec kube-master docker load -i /calico-node.tar
+	docker exec kube-node-1 docker load -i /calico-node.tar
+	docker exec kube-node-2 docker load -i /calico-node.tar
+	kubectl apply -f  https://docs.projectcalico.org/v3.3/getting-started/kubernetes/installation/hosted/calicoctl.yaml
```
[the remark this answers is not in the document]
lwr20: It's calicoctl so that its available to the tests to use.  One of the recommended ways to install calicoctl these days is using that manifest.  Then when you want to run a calicoctl command, you just `kubectl exec` in that pod.  

The manifest picks up the data store config from the cluster itself.

tests/k8st/dind-cluster.sh
```diff
@@ -0,0 +1,2336 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
lwr20: LOL @neiljerram told me to do exactly that to avoid things in our build changing under our feet.

And no, its not an exact copy of the script from the repo you reference above.  

OK - in that case the other possibility is to fork the kubeadm-dind-cluster repo into tigera, make our changes and then reference that here.

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -0,0 +1,217 @@
...
+        self.tearDown()
+
+        # # Setup external node: use privileged mode for setting routes
+        subprocess.check_call("docker run -d "
+                              "--privileged "
+                              "--name kube-node-extra "
+                              "--network kubeadm-dind-net "
+                              "mirantis/kubeadm-dind-cluster:v1.10",
+                              shell=True)
+        # Install bird on extra node
```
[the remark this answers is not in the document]
lwr20: In general, I agree that there's lots of code in that test that needs to be encapsulated and put somewhere for other things to use.  e.g. starting pods and a service.  

But this particular instance?  I don't recall *any* other test suite where starting BIRD is useful (apart from the openstack tests).

OTOH, its a cheap change to make and might well make the test more readable.

calico_test/Dockerfile.amd64.calico_test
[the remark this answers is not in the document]
neiljerram: I really think we need a comment above that COPY statement, explaining it.

tests/k8st/dind-cluster.sh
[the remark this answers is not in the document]
lwr20: Either way, we're *not* using the script from upstream unmodified.  

So we need to fork `kubeadm-dind-cluster` somewhere.  And it ought to be into this org, no?

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -0,0 +1,217 @@
...
+template bgp bgp_template {
+  debug { states };
+  description "Connection to BGP peer";
+  local as 64512;
+  multihop;
+  gateway recursive; # This should be the default, but just in case.
+  import all;        # Import all routes, since we don't know what the upstream
+                     # topology is and therefore have to trust the ToR/RR.
+  export all;
+  source address 10.192.0.5;  # The local address we use for the TCP connection
```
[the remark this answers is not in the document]
lwr20: Yes.

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -0,0 +1,217 @@
...
+                              shell=True)
+        # Install bird on extra node
+        subprocess.check_call("docker exec kube-node-extra apt update", shell=True)
+        subprocess.check_call("docker exec kube-node-extra apt install -y bird", shell=True)
+        subprocess.check_call("docker exec kube-node-extra mkdir /run/bird", shell=True)
+        with open('bird.conf', 'w') as birdconfig:
+            birdconfig.write(bird_conf)
+        subprocess.check_call("docker cp bird.conf kube-node-extra:/etc/bird/bird.conf", shell=True)
+        subprocess.check_call("rm bird.conf", shell=True)
+        subprocess.check_call("docker exec kube-node-extra service bird restart", shell=True)
```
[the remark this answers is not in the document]
lwr20: > Also, a key point here is that the extra node must not be a member of the k8s cluster, and in particular not run kube-proxy. Is that the case with this setup?
Yes, it absolutely is.

And in fact I was worried that by using calico-node, we might end up with more integration with the cluster than we were expecting (e.g. bird config).  But happy to do that if you tell me its safe to do.  

The way it is now, we can see *exactly* what's going into that node right here in the test.  Moving that stuff to a Dockerfile moves that config out of this test, which might be OK if we were reusing that container somewhere else, but I don't see that we will (at least for now)

tests/k8st/tests/test_simple.py
```diff
@@ -0,0 +1,198 @@
...
+            template=template)
+        # Instantiate the deployment object
+        nginx_deployment = client.ExtensionsV1beta1Deployment(
+            api_version="extensions/v1beta1",
+            kind="Deployment",
+            metadata=client.V1ObjectMeta(name="nginx"),
+            spec=spec)
+        api_response = client.ExtensionsV1beta1Api().create_namespaced_deployment(
+            body=nginx_deployment,
+            namespace="policy-demo")
```
neiljerram: It's hard to read what's going on with the calls above.

- If it's just "kubectl run nginx --namespace policy-demo", can we just do that instead?

- Alternatively, move into a subroutine called run_nginx or something.

tests/k8st/tests/test_simple.py
```diff
@@ -0,0 +1,198 @@
...
+            ),
+            spec={
+                "ports": [{"port": 80}],
+                "selector": {"app": "nginx"},
+            }
+        )
+        api_response = cluster.create_namespaced_service(
+            body=service,
+            namespace="policy-demo",
+        )
```
neiljerram: Here too, it feels like the kubectl command would be clearer.

tests/k8st/tests/test_simple.py
```diff
@@ -0,0 +1,198 @@
...
+        #
+        # # Check we cannot talk to service as 'cant-access'
+        # # Check we cannot talk to service
+        # succeeded = False
+        # for i in range(5):
+        #     sleep(1)
+        #     if not self.check_connected("cant-access"):
+        #         succeeded = True
+        #         break
+        # assert succeeded is True
```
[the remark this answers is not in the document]
neiljerram: With this commenting out, is there any testing of Calico policy?

tests/k8st/tests/test_simple.py
```diff
@@ -0,0 +1,198 @@
...
+        try:
+            subprocess.check_call("kubectl run "
+                                  "--namespace=policy-demo "
+                                  "%s "
+                                  "--restart Never "
+                                  "--rm -i "
+                                  "--image busybox "
+                                  "--command /bin/wget "
+                                  "-- -q --timeout=1 nginx" % name,
+                                  shell=True)
```
neiljerram: That's the way to do it! :smile:

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -0,0 +1,217 @@
...
+  asNumber: 64512
+EOF
+""", shell=True)
+
+    def tearDown(self):
+        deleted_stuff = False
+        try:
+            subprocess.check_call("docker rm -f kube-node-extra", shell=True)
+        except subprocess.CalledProcessError:
+            deleted_stuff = True
```
neiljerram: Why is deleted_stuff `True` when these deletion calls *fail*?

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -0,0 +1,217 @@
...
+        # # Establish BGPPeer from cluster nodes to node-extra using calicoctl
+        subprocess.check_call("""kubectl exec -i -n kube-system calicoctl -- /calicoctl apply -f - << EOF
+apiVersion: projectcalico.org/v3
+kind: BGPPeer
+metadata:
+  name: node-extra.peer
+spec:
+  peerIP: 10.192.0.5
+  asNumber: 64512
+EOF
```
neiljerram: Can that `kubectl ... calicoctl ...` be collapsed into a kubectl operation on a CRD?

(Genuinely don't know here; I might be talking rubbish, but I thought we supported something like that.)

tests/k8st/tests/test_bgp_advert.py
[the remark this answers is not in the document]
neiljerram: OK, I'm persuaded for now.  If we subroutine this and start using it multiple times, it probably will be worth doing the `apt update` and `apt install -y bird` once upfront; but no need to do that now.

Re using calico-node I take your point.  FYI I think there are 3 relevant areas of function, if we used calico-node.
1. Whether the node/host is a k8s member and has kube-proxy.  This would completely undermine the test concept.  It depends entirely on the kubeadm-dind-cluster setup, and you seem confident that the right thing is happening - so good.
2. The BIRD, BIRD config and confd logic inside calico-node.  IMO it would definitely be a _feature_ for us to be using those, not a concern; and I'm confident that they don't in any way undermine what we're trying to test.
3. The Felix inside calico-node.  Theoretically I think this is fine, but I do feel a bit of FUD about whether running Felix might somehow interact with routing and possibly undermine the test intention.

Perhaps we could usefully do some more polishing to facilitate using calico/node in this kind of context; but no need to do that in this particular ST.

tests/k8st/tests/test_bgp_advert.py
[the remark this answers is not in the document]
neiljerram: I think I favour encapsulation too.

tests/k8st/tests/test_simple.py
lwr20: Yeah - in general, the kubernetes API calls are all much nastier than simply running the kubectl equivalent :)

tests/k8st/tests/test_simple.py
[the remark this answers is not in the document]
lwr20: TBH, this is irrelevant to the task I'm currently working on (testing BGP stuff).  This test should be removed in its entirety for now.

Makefile
```diff
@@ -451,6 +453,47 @@ st-checks:
...
+	    --net host \
+        $(TEST_CONTAINER_NAME) \
+	    sh -c 'cp /root/.kubeadm-dind-cluster/kubectl /bin/kubectl && ls -ltr /bin/kubectl && which kubectl && cd /code/tests/k8st && nosetests -v --with-xunit --xunit-file="/code/report/k8s-tests.xml" --with-timer'
+	$(MAKE) k8s-stop
+
+.PHONY: k8s-start
+## Start k8s cluster
+k8s-start: tests/k8st/dind-cluster.sh calico-node.tar
+	CNI_PLUGIN=custom POD_CIDR=192.168.0.0/16 SKIP_SNAPSHOT=y tests/k8st/dind-cluster.sh up
+	kubectl apply -f  https://docs.projectcalico.org/master/getting-started/kubernetes/installation/hosted/etcd.yaml
```
caseydavenport: Hm, all these manifest links have a hardcoded `master` set.

How are we planning on supporting multiple release branches here? Can we just fill in the `master` bit with the correct `vX.Y` based on the branch?

Makefile
```diff
@@ -451,6 +453,47 @@ st-checks:
...
+	docker run \
+	    -v $(CURDIR):/code \
+	    -v /var/run/docker.sock:/var/run/docker.sock \
+	    -v /home/$(USER)/.kube/config:/root/.kube/config \
+	    -v /home/$(USER)/.kubeadm-dind-cluster:/root/.kubeadm-dind-cluster \
+	    --privileged \
+	    --net host \
+        $(TEST_CONTAINER_NAME) \
+	    sh -c 'cp /root/.kubeadm-dind-cluster/kubectl /bin/kubectl && ls -ltr /bin/kubectl && which kubectl && cd /code/tests/k8st && nosetests -v --with-xunit --xunit-file="/code/report/k8s-tests.xml" --with-timer'
+	$(MAKE) k8s-stop
```
caseydavenport: Do we want to do this as part of the test target?

It would be really useful if a test fails to be able to poke around the cluster afterwards.

tests/k8st/dind-cluster.sh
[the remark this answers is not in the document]
caseydavenport: @neiljerram I think we can remove this checked-in copy now.

tests/k8st/test_base.py
```diff
@@ -0,0 +1,183 @@
...
+        self.cluster.create_namespace(client.V1Namespace(metadata=client.V1ObjectMeta(name=ns_name)))
+
+    def create_service(self, image, name, ns, port, replicas=2):
+        cluster = self.k8s_client()
+        cluster.create_namespace(client.V1Namespace(metadata=client.V1ObjectMeta(name=ns)))
+        container = client.V1Container(
+            name=name,
+            image=image,
+            ports=[client.V1ContainerPort(container_port=port)])
+        # Create and configure a spec section
```
caseydavenport: We're going to want the ability to control deployments / services / endpoints independently of each other, I think.

I think it would be good to split this out into separate functions.

caseydavenport: Could add a convenience wrapper around them if needed.

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -0,0 +1,160 @@
...
+bird_conf = """
+router id 10.192.0.5;
+
+# Configure synchronization between routing tables and kernel.
+protocol kernel {
+  learn;             # Learn all alien routes from the kernel
+  persist;           # Don't remove routes on bird shutdown
+  scan time 2;       # Scan kernel routing table every 2 seconds
+  import all;
+        export all;
```
caseydavenport: indentation is wonky here.

tests/k8st/tests/test_simple.py
```diff
@@ -0,0 +1,134 @@
...
+            body=policy,
+            namespace="policy-demo",
+        )
+        _log.debug("Allow policy created.")
+
+        # Check we can talk to service as 'access'
+        retry_until_success(self.can_connect, retries=10, wait_time=1, function_args=["access"])
+
+        # Check we cannot talk to service as 'cant-access'
+        retry_until_success(self.cannot_connect, retries=10, wait_time=1, function_args=["access"])
```
caseydavenport: Should this say `function-args=["cant-access"]`

Makefile
```diff
@@ -451,6 +453,47 @@ st-checks:
...
+	CNI_PLUGIN=custom POD_CIDR=192.168.0.0/16 SKIP_SNAPSHOT=y tests/k8st/dind-cluster.sh up
+	kubectl apply -f  https://docs.projectcalico.org/master/getting-started/kubernetes/installation/hosted/etcd.yaml
+	docker cp calico-node.tar kube-master:/calico-node.tar
+	docker cp calico-node.tar kube-node-1:/calico-node.tar
+	docker cp calico-node.tar kube-node-2:/calico-node.tar
+	docker exec kube-master docker load -i /calico-node.tar
+	docker exec kube-node-1 docker load -i /calico-node.tar
+	docker exec kube-node-2 docker load -i /calico-node.tar
+	wget https://docs.projectcalico.org/master/getting-started/kubernetes/installation/hosted/calico.yaml -O - | \
+		sed s,quay.io/calico/node:master,$(BUILD_IMAGE):latest-$(ARCH), | \
```
caseydavenport: I wouldn't rely on the `quay` repo here - this is likely to change in a future release. If we could do a sed that handles varying repository names that would be better.

Makefile
neiljerram: I think mainline is for the test to run and clean up after itself.  It's easy enough for a developer to comment that line when they want to poke around.  WDYT?

tests/k8st/tests/test_simple.py
neiljerram: Hmm, well it's very odd that this test can be passing as is!

Makefile
caseydavenport: I think we have a DEBUG_FAILURES flag defined for our other e2es - can we base it off that?

I'm not a huge fan of workflows that rely on people knowing which bits of code to comment out.

tests/k8st/tests/test_simple.py
neiljerram: Fixed now.  (After a long time understanding how it could be passing!)

tests/k8st/utils/utils.py
```diff
@@ -0,0 +1,103 @@
...
+    subprocess.check_call("docker exec %s apt install -y bird" % name, shell=True)
+    subprocess.check_call("docker exec %s mkdir /run/bird" % name, shell=True)
+    with open('bird.conf', 'w') as birdconfig:
+        birdconfig.write(config)
+    subprocess.check_call("docker cp bird.conf %s:/etc/bird/bird.conf" % name, shell=True)
+    subprocess.check_call("rm bird.conf", shell=True)
+    subprocess.check_call("docker exec %s service bird restart" % name, shell=True)
+
+
+def retry_until_success(fun,
```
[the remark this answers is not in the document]
neiljerram: Yeah, but it's a subtly different version.  I had a go at merging them, but it became a rabbit hole.  Still might be a good thing, but can be a follow up.

Makefile
```diff
@@ -115,6 +115,7 @@ BIRD_IMAGE ?= calico/bird:$(BIRD_VER)-$(ARCH)
 # Versions and locations of dependencies used in tests.
 CALICOCTL_VER?=master
 CNI_VER?=master
+MANIFEST_VER?=master
```
caseydavenport: Doesn't look like this variable is used anywhere?

tests/k8st/test_base.py
```diff
@@ -0,0 +1,182 @@
...
+
+        for pod in pods.items:
+            logger.info("%s\t%s\t%s", pod.metadata.name, pod.metadata.namespace, pod.status.phase)
+            assert pod.status.phase == 'Running'
+
+    def create_namespace(self, ns_name):
+        self.cluster.create_namespace(client.V1Namespace(metadata=client.V1ObjectMeta(name=ns_name)))
+
+    def create_service(self, image, name, ns, port, replicas=2):
+        cluster = self.k8s_client()
```
caseydavenport: Wondering if we should create the namespace here - feels wrong to do that implicitly when creating a service.

tests/k8st/test_base.py
```diff
@@ -0,0 +1,182 @@
...
+        pods = self.cluster.list_namespaced_pod(ns)
+
+        for pod in pods.items:
+            logger.info("%s\t%s\t%s", pod.metadata.name, pod.metadata.namespace, pod.status.phase)
+            assert pod.status.phase == 'Running'
+
+    def create_namespace(self, ns_name):
+        self.cluster.create_namespace(client.V1Namespace(metadata=client.V1ObjectMeta(name=ns_name)))
+
+    def create_service(self, image, name, ns, port, replicas=2):
```
caseydavenport: We probably want to be able to specify the externalTrafficPolicy here

tests/k8st/test_base.py
```diff
@@ -0,0 +1,182 @@
...
+    def check_calico_version(cls):
+        config.load_kube_config(os.environ.get('KUBECONFIG'))
+        api = client.AppsV1Api(client.ApiClient())
+        node_ds = api.read_namespaced_daemon_set("calico-node", "kube-system", exact=True, export=True)
+        for container in node_ds.spec.template.spec.containers:
+            if container.name == "calico-node":
+                if container.image != "calico/node:latest-amd64":
+                    container.image = "calico/node:latest-amd64"
+                    api.replace_namespaced_daemon_set("calico-node", "kube-system", node_ds)
+                    time.sleep(3)
```
caseydavenport: Is this a stray sleep? We do some retries below which should handle this?

tests/k8st/test_base.py
neiljerram: I believe Brian has a markup ready for that.

Makefile
neiljerram: This change is in the kubeadm-dind-cluster repo now - but I will improve it there.

tests/k8st/test_base.py
neiljerram: I guess the idea is to ensure that the replacement has _started_.  Would that make sense?

tests/k8st/utils/utils.py
```diff
@@ -0,0 +1,124 @@
...
+    # Setup external node: use privileged mode for setting routes
+    run("docker run -d "
+        "--privileged "
+        "--name %s "
+        "--network kubeadm-dind-net "
+        "mirantis/kubeadm-dind-cluster:tar" % name)
+
+    # It may take a while for the image to download and the container
+    # to start running.
+    while True:
```
caseydavenport: Might we want to limit this to some amount of time? 

I guess on semaphore we'll eventually time out.

caseydavenport:
@neiljerram hitting some space issues on semaphore it looks like. 

```
Successfully built 339042cb4555
Successfully tagged calico/test:latest-amd64
touch calico_test.created
docker tag mirantis/kubeadm-dind-cluster:v1.12 mirantis/kubeadm-dind-cluster:tar
docker save --output kubeadm-dind-cluster.tar mirantis/kubeadm-dind-cluster:tar
write .docker_temp_525723040: no space left on device
Makefile:489: recipe for target 'kubeadm-dind-cluster.tar' failed
```

tests/k8st/utils/utils.py
neiljerram: Yes, I think relying on the semaphore time out may make sense here.  (Sometimes networking from semaphore is very slow; I'm not sure how to put a reasonable bound on that.)

PR #117, opened 2018-11-30
add first mainline test for service ip advertisement

briansan:


tests/k8st/test_base.py
```diff
@@ -182,3 +183,32 @@ def is_it_gone_yet(res_name, res_type):
...
+    def get_routes(self):
+        return run("docker exec kube-node-extra ip r")
+
+    def update_ds_env(self, ds, ns, k, v):
+        config.load_kube_config(os.environ.get('KUBECONFIG'))
+        api = client.AppsV1Api(client.ApiClient())
+        node_ds = api.read_namespaced_daemon_set(ds, ns, exact=True, export=True)
+        for container in node_ds.spec.template.spec.containers:
+            if container.name == ds:
+                route_env_present = False
```
caseydavenport: This variable name doesn't make sense any more

tests/k8st/test_base.py
```diff
@@ -182,3 +183,32 @@ def is_it_gone_yet(res_name, res_type):
 
     class StillThere(Exception):
         pass
+
+    def get_routes(self):
+        return run("docker exec kube-node-extra ip r")
+
+    def update_ds_env(self, ds, ns, k, v):
```
caseydavenport: Not for this PR, but it might be nice to support a list of env vars to set if we ever want to update multiple things.

Just because it takes a long time to wait for each rolling update to occur.

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -152,20 +125,46 @@ def tearDown(self):
...
+        - Scale the Local NP service so it is running on multiple nodes, assert ECMP routing, source IP is preserved.
+        - Delete both services, assert only cluster CIDR route is advertised.
         """
+        # Assert that a route to the service IP range is present
+        retry_until_success(lambda: self.assertIn("10.96.0.0/12", self.get_routes()), retries=5, wait_time=1)
+
+        # Create nginx deployment and service
+        self.create_service("nginx:1.7.9", "nginx", "bgp-test", 80)
+        self.create_service("nginx:1.7.9", "nginx2", "bgp-test", 80, traffic_policy="Cluster")
+        sleep(3) # TODO: hack
```
caseydavenport: What's up with this TODO hack?

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -152,20 +125,46 @@ def tearDown(self):
...
+        - Create a local LoadBalancer service with clusterIP = None, assert no change.
+        - Scale the Local NP service so it is running on multiple nodes, assert ECMP routing, source IP is preserved.
+        - Delete both services, assert only cluster CIDR route is advertised.
         """
+        # Assert that a route to the service IP range is present
+        retry_until_success(lambda: self.assertIn("10.96.0.0/12", self.get_routes()), retries=5, wait_time=1)
+
+        # Create nginx deployment and service
+        self.create_service("nginx:1.7.9", "nginx", "bgp-test", 80)
+        self.create_service("nginx:1.7.9", "nginx2", "bgp-test", 80, traffic_policy="Cluster")
```
caseydavenport: nit: I'd call these `nginx-local` and `nginx-cluster` for easier time debugging if things go wrong.

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -152,20 +125,46 @@ def tearDown(self):
...
+        clusterIP1 = self.get_svc_cluster_ip("nginx", "bgp-test")
+        clusterIP2 = self.get_svc_cluster_ip("nginx2", "bgp-test")
+
+        # Assert that both nginx service can be curled from the external node
+        self.do_external_curl("kube-node-extra", clusterIP1)
+        self.do_external_curl("kube-node-extra", clusterIP2)
+
+        # Assert that local clusterIP is an advertised route and cluster clusterIP is not
+        retry_until_success(lambda: self.assertIn(clusterIP1, self.get_routes()), retries=5, wait_time=1)
+        retry_until_success(lambda: self.assertNotIn(clusterIP2, self.get_routes()), retries=5, wait_time=1)
```
caseydavenport: I don't think you need the lambda here - it should work, but something like this would be just as good:

```
retry_until_success(self.assertNotIn, function_args=[clusterIP2, self.get_routes()], retries=5, wait_time=1)
```

PR #118, opened 2018-11-30
add in remaining set of mainline tests for service ip advertisement

briansan:


tests/k8st/tests/test_bgp_advert.py
```diff
@@ -165,7 +165,42 @@ def test_mainline(self):
...
+    protocol: TCP
+    source: { nets: ["10.192.0.5/32"] }
+    destination: { ports: ["80"] }
+EOF
+""")
+
+        # Connectivity to nginx-local should always succeed
+        self.do_external_curl("kube-node-extra", local_svc_ip)
+
+        # Connectivity to nginx-cluster will rarely succeed, try to curl 10 times
```
caseydavenport: Would be worth adding a "why" to this comment, for future reviewers.

e.g. "will rarely succeed because it is load-balanced across all nodes. When the traffic hits a node that doesn't host one of the service's pod, it will be re-routed to another node and SNAT will cause the policy to drop the traffic"

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -165,7 +165,42 @@ def test_mainline(self):
...
+        try:
+          for i in range(10):
+            self.do_external_curl("kube-node-extra", cluster_svc_ip)
+          self.fail("external node should not be able to access cluster svc")
```
caseydavenport: "should not be able to _consistently_ access..."

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -165,7 +165,42 @@ def test_mainline(self):
...
+        except subprocess.CalledProcessError:
+          pass
+
+        # Scale the local_svc to 2 replicas
+        self.scale_deployment(local_svc, self.ns, 2)
+        matchStr = local_svc_ip + " proto bird \n\tnexthop via 10.192.0.3  dev eth0 weight 1\n\tnexthop via 10.192.0.4  dev eth0 weight 1"
```
caseydavenport: It would be really nice to wrap this up into a helper function - `assert_ecmp_routes` or something like that.

e.g. `self.assert_ecmp_routes(dst, via=[])`

Then, we could use it more generally to assert routes are balanced across multiple vias.

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -165,7 +165,42 @@ def test_mainline(self):
...
+        retry_until_success(lambda: self.assertIn(matchStr, self.get_routes()))
+
```
caseydavenport: Let's also assert that accessing the local svc still works - e.g. loop through 10 times and make sure each access is successful.

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -165,7 +165,42 @@ def test_mainline(self):
...
+  ingress:
+  - action: Allow
+    protocol: TCP
+    source: { nets: ["10.192.0.5/32"] }
+    destination: { ports: ["80"] }
+EOF
+""")
+
+        # Connectivity to nginx-local should always succeed
+        self.do_external_curl("kube-node-extra", local_svc_ip)
```
caseydavenport: I think we want to retry this 10 times or so as well, to make sure we don't just happen to land on the right node / pod.

This one should consistently succeed, whereas the one below should fail sometimes, succeed sometimes.

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -165,7 +165,42 @@ def test_mainline(self):
...
+
+        # Connectivity to nginx-cluster will rarely succeed, try to curl 10 times
+        try:
+          for i in range(10):
+            self.do_external_curl("kube-node-extra", cluster_svc_ip)
+          self.fail("external node should not be able to access cluster svc")
+        except subprocess.CalledProcessError:
+          pass
+
+        # Scale the local_svc to 2 replicas
```
caseydavenport: I'd recommend doing 3-4 replicas, just to ensure there's a replica on more than one node (I'd be worried about two ending up on the same node for some reason).

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -165,7 +165,42 @@ def test_mainline(self):
         retry_until_success(lambda: self.assertIn(local_svc_ip, self.get_routes()))
         retry_until_success(lambda: self.assertNotIn(cluster_svc_ip, self.get_routes()))
 
-        # Delete nginx service
+        # Create a network policy that only accepts traffic from the external node
+        run("""kubectl exec -i -n kube-system calicoctl -- /calicoctl apply -f - << EOF
```
caseydavenport: This is fine, but it might be easier to use a k8s NP rather than a Calico one - that way, it gets automatically deleted when the namespace is deleted.

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -142,7 +158,7 @@ def test_mainline(self):
         - Scale the Local NP service so it is running on multiple nodes, assert ECMP routing, source IP is preserved.
         - Delete both services, assert only cluster CIDR route is advertised.
         """
-        # Assert that a route to the service IP range is present
+        # Assert that a route to the service IP range is present.
```
caseydavenport: Nice.

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -153,52 +169,51 @@ def test_mainline(self):
...
+        # Get clusterIPs.
         local_svc_ip = self.get_svc_cluster_ip(local_svc, self.ns)
         cluster_svc_ip = self.get_svc_cluster_ip(cluster_svc, self.ns)
 
-        # Assert that both nginx service can be curled from the external node
-        retry_until_success(self.do_external_curl, function_args=["kube-node-extra", local_svc_ip])
-        retry_until_success(self.do_external_curl, function_args=["kube-node-extra", cluster_svc_ip])
+        # Assert that both nginx service can be curled from the external node.
+        self.do_external_curl("kube-node-extra", local_svc_ip)
+        self.do_external_curl( "kube-node-extra", cluster_svc_ip)
```
caseydavenport: extra space here

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -129,8 +129,24 @@ def tearDown(self):
     def get_svc_cluster_ip(self, svc, ns):
         return run("kubectl get svc %s -n %s -o json | jq -r .spec.clusterIP" % (svc, ns)).strip()
 
-    def do_external_curl(self, container, hostname):
-        return run("docker exec %s curl --connect-timeout 2 -m 3 %s" % (container, hostname))
+    def do_external_curl(self, container, hostname, succeed=True):
+        out = "failed"
+        try:
+          for i in range(10):
```
caseydavenport: I'm.. not a fan of adding the loop into `do_external_curl`.

I think there are a lot of situations where we don't want this, and it's better to allow the calling code to make the decision as to whether or not it wants to retry at all, and if so how many times.

Can we revert this change and just loop over the call within the test code itself?

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -129,43 +129,93 @@ def tearDown(self):
...
+            cmd = "docker exec %s curl --connect-timeout 2 -m 3 %s" % (container, hostname)
+            out = retry_until_success(run, function_args=[cmd]) if succeed else run(cmd)
+          if not succeed:
+            self.fail(container + " should not be able to consistently access " + hostname)
+        except subprocess.CalledProcessError:
+          if succeed:
+            raise
+        return out
+
+    def assertECMPRoutes(self, dst, via=["10.192.0.3", "10.192.0.4"]):
```
caseydavenport: golang seeping in here.. should be `assert_ecmp_routes` :)

briansan: i did this intentionally since all the assert methods in the python `unittest.TestCase` are camelCased, but maybe it's better to stick to the 🐍

caseydavenport:
@briansan if you give this a squash I'll merge

PR #119, opened 2018-12-01
Add test for many services, endpoint scaling

caseydavenport:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/tests/k8st/test_base.py
+++ b/tests/k8st/test_base.py
@@ -35,15 +35,13 @@ class TestBase(TestCase):
     """
     Base class for test-wide methods.
     """
-    @classmethod
-    def setUpClass(cls):
-        cls.check_calico_version()
 
     def setUp(self):
         """
         Clean up before every test.
         """
         self.cluster = self.k8s_client()
+        self.check_calico_version()
 
         # Log a newline to ensure that the first log appears on its own line.
         logger.info("")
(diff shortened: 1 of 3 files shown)


tests/k8st/test_base.py
```diff
@@ -152,6 +152,27 @@ def create_service(self, image, name, ns, port, replicas=1, svc_type="NodePort",
         )
         logger.debug("Service created. status='%s'" % str(api_response.status))
 
+    def create_additional_service(self, name, app, ns, port, svc_type="NodePort", traffic_policy="Local", cluster_ip=None):
```
briansan: This seems like an exact duplicate to L135-153. Could we refactor `create_service` to also call `create_additional_service` or unless there some divergent logic that I missed?

caseydavenport: Yep, done!

tests/k8st/test_base.py
```diff
@@ -35,15 +35,13 @@ class TestBase(TestCase):
     """
     Base class for test-wide methods.
     """
-    @classmethod
```
caseydavenport: Unless I'm missing something, I think this was broken, since it was trying to use fields that weren't instantiated until `setUp` - namely, the k8s client.

PR #124, opened 2018-12-06
Better diags on failed tests

caseydavenport:
## Description

This PR adds, hopefully, some additional diagnostics to k8s fv tests.

namely:
- Waits for deployments to finish rolling out before carrying on with test.
- Adds some basic diags collection to be printed to logs on failure.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/tests/k8st/utils/utils.py
+++ b/tests/k8st/utils/utils.py
@@ -19,6 +19,23 @@ import time
 
 _log = logging.getLogger(__name__)
 
+# Helps with printing diags after a test.
+class DiagsCollector(object):
+    def __enter__(self):
+        pass
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        # Print out diagnostics for the test. These will go to screen
+        # on test failure.
+        _log.info("===================================================")
+        _log.info("============= COLLECTING DIAGS FOR TEST ===========")
+        _log.info("===================================================")
+        run("kubectl get deployments,pods,svc,endpoints --all-namespaces -o wide")
+        run("kubectl logs -n kube-system -l k8s-app=calico-node")
+        _log.info("===================================================")
+        _log.info("============= COLLECTED DIAGS FOR TEST ============")
+        _log.info("===================================================")
+
 
 def start_external_node_with_bgp(name, config):
     # Check how much disk space we have.
--- a/tests/k8st/test_base.py
+++ b/tests/k8st/test_base.py
@@ -136,6 +136,14 @@ class TestBase(TestCase):
         # with "app": <name>; i.e. those just created above.
         self.create_service(name, name, ns, port, svc_type, traffic_policy)
 
+    def wait_for_deployment(self, name, ns):
+        """
+        Waits for the given deployment to have the desired number of replicas.
+        """
+        logger.info("Checking status for deployment %s/%s" % (ns, name))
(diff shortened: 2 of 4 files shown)


tests/k8st/utils/utils.py
```diff
@@ -19,6 +19,23 @@
...
+        pass
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        # Print out diagnostics for the test. These will go to screen
+        # on test failure.
+        _log.info("===================================================")
+        _log.info("============= COLLECTING DIAGS FOR TEST ===========")
+        _log.info("===================================================")
+        run("kubectl get deployments,pods,svc,endpoints --all-namespaces -o wide")
+        run("kubectl logs -n kube-system -l k8s-app=calico-node")
```
briansan: While this command is nice and succinct, it looks like this will dump the logs in a fashion that's a little difficult to delimit them. 
Do you think we could execute something like `kubectl get pods -nkube-system | grep calico-node | awk '{ printf $1"\n" }'`, iterate over the result.split(), and print a header with the pod name + the log dump for each?

tests/k8st/utils/utils.py
```diff
@@ -19,6 +19,23 @@
 
 _log = logging.getLogger(__name__)
 
+# Helps with printing diags after a test.
+class DiagsCollector(object):
```
briansan: I think it would be good to define an `__init__` with some args that customize what the Diag outputs.
As an example, there could be like a `dump_routes` variable to print out `docker exec kube-node-extra ip r` for service advertisement tests

caseydavenport:
> Looks nice. Is this now the complete fix as well? I.e. does the waiting get rid of the flake?

It's hard to say, I need to get quite a few more test runs in to see if we still hit the same flakes.

tests/k8st/utils/utils.py
caseydavenport: It _should_ be easy enough to find the beginning of each - we include some standard logs on startup that should make it clear where one log file begins, and which node it is for.

briansan:
```
2018-12-06 21:51:20.987 [INFO][45] route_table.go 189: Queueing a resync of routing table. ipVersion=0x4
2018-12-06 21:51:20.989 [INFO][45] int_dataplane.go 765: Finished applying updates to dataplane. msecToApply=2.1740790000000003
2018-12-06 21:51:22.054 [INFO][45] health.go 150: Overall health summary=&health.HealthReport{Live:true, Ready:true}
2018-12-06 20:37:48.793 [INFO][8] startup.go 244: Early log level set to info
2018-12-06 20:37:48.793 [INFO][8] startup.go 264: Using HOSTNAME environment (lowercase) for node name
2018-12-06 20:37:48.794 [INFO][8] startup.go 272: Determined node name: kube-master
```
good enough for me :+1:

tests/k8st/utils/utils.py
caseydavenport: Yeah, we could add some arguments to this, but for the information I was looking for right now I didn't need any so I didn't add any :D 

Once there's a compelling reasons to pass arguments we can add them.

Dumping routes on the remote node already shows up in the log output, but adding confd template dumps is on my list of things we could do for sure.

PR #125, opened 2018-12-06
Additional assertion on BGP peering state

caseydavenport:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/tests/st/bgp/test_single_route_reflector.py
+++ b/tests/st/bgp/test_single_route_reflector.py
@@ -17,7 +17,7 @@ from unittest import skip
 from tests.st.test_base import TestBase
 from tests.st.utils.docker_host import DockerHost, CLUSTER_STORE_DOCKER_OPTIONS
 from tests.st.utils.route_reflector import RouteReflectorCluster
-from tests.st.utils.utils import update_bgp_config
+from tests.st.utils.utils import update_bgp_config, check_bird_status, retry_until_success
 
 from .peer import create_bgp_peer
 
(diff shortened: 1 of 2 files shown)


neiljerram:
```AssertionError: '10.110.108.204 proto bird \n\tnexthop via 10.192.0.3  dev eth0 weight 1\n\tnexthop via 10.192.0.4  dev eth0 weight 1' not found in 'default via 10.192.0.1 dev eth0 \n10.96.0.0/12 proto bird \n\tnexthop via 10.192.0.2  dev eth0 weight 1\n\tnexthop via 10.192.0.3  dev eth0 weight 1\n\tnexthop via 10.192.0.4  dev eth0 weight 1\n10.110.108.204 via 10.192.0.3 dev eth0 proto bird \n10.192.0.0/24 dev eth0 proto kernel scope link src 10.192.0.5 \n192.168.135.128/26 via 10.192.0.3 dev eth0 proto bird \n192.168.169.128/26 via 10.192.0.4 dev eth0 proto bird \n192.168.221.192/26 via 10.192.0.2 dev eth0 proto bird \n'
```
The test code is expecting ECMP routes for 10.110.108.204, but apparently there is only a single path programmed.

caseydavenport:
> The test code is expecting ECMP routes for 10.110.108.204, but apparently there is only a single path programmed.

Nice, this is one of the flakes we saw yesterday as well. I think we need additional diags to find the cause - will try to get a PR which spits out BIRD config, etc., on failure.

caseydavenport:
Hoping that diags from here can help: https://github.com/projectcalico/node/pull/124

PR #114, opened 2018-11-29
Allow more time for convergence after BGPPeer config to RRs

neiljerram:
I think we've been hitting a BIRD backoff of 30s because of connection
collision; and then this test's allowance of 25s isn't enough.  Details
for a particular
case (https://semaphoreci.com/calico/node-private/branches/confd-start-of-day/builds/1)
are:

Diags collection began at 16:15:15.851326

Ping attempts happened between 16:14:50.919387 and 16:15:14.415218

Problem was workload1_hihxkM failing to ping 192.168.190.0

BGPPeers were configured by 16:14:50.708608

workload1_hihxkM is on host1

At 16:15:30.620864, ip r on host1 includes 192.168.190.0/26 via 172.17.0.4 dev eth0 proto bird

At 16:15:24.877942, ip r on host3 includes 192.168.190.0 dev cali0abb0456e76 scope link

host1's BIRD log, between 16:14:48 (when the BGPPeers were configured) and the end of the test, was:
2018-11-29_16:14:48.69737 bird: Reconfigured
2018-11-29_16:14:50.11571 bird: Node_172_17_0_6: Connected to table master
2018-11-29_16:14:50.11572 bird: Node_172_17_0_6: State changed to feed
2018-11-29_16:14:50.11590 bird: Node_172_17_0_6: State changed to up
2018-11-29_16:14:50.11665 bird: Node_172_17_0_5: Connected to table master
2018-11-29_16:14:50.11665 bird: Node_172_17_0_5: State changed to feed
2018-11-29_16:14:50.11668 bird: Node_172_17_0_5: State changed to up
2018-11-29_16:14:50.11761 bird: Node_172_17_0_5: State changed to start
2018-11-29_16:14:52.12290 bird: Node_172_17_0_5: State changed to feed
2018-11-29_16:14:52.12291 bird: Node_172_17_0_5: Received: Connection collision resolution
2018-11-29_16:14:52.12291 bird: Node_172_17_0_5: State changed to stop
2018-11-29_16:14:52.12302 bird: Node_172_17_0_5: State changed to down
2018-11-29_16:14:52.12302 bird: Node_172_17_0_5: Starting
2018-11-29_16:14:52.12302 bird: Node_172_17_0_5: State changed to start
2018-11-29_16:14:54.12649 bird: Node_172_17_0_5: Connected to table master
2018-11-29_16:14:54.12650 bird: Node_172_17_0_5: State changed to feed
[...]

PR #126, opened 2018-12-07
Add assertions that BGP is established

caseydavenport:
## Description

Not sure if this is what is going on here, but thought it worth a shot since we've seen this failure a few times recently: https://semaphoreci.com/calico/node/branches/master/builds/363

We know that sometimes BGP peerings can be racy if they try to initiate at the same time and hit a conflict, causing a longer-than-expected delay in getting routes shared.

This should hopefully wait for BGP to establish before asserting reachability. 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/tests/st/bgp/test_route_reflector_cluster.py
+++ b/tests/st/bgp/test_route_reflector_cluster.py
@@ -19,7 +19,7 @@ from tests.st.utils.docker_host import DockerHost, CLUSTER_STORE_DOCKER_OPTIONS
 from tests.st.utils.route_reflector import RouteReflectorCluster
 
 from .peer import create_bgp_peer
-from tests.st.utils.utils import update_bgp_config
+from tests.st.utils.utils import update_bgp_config, retry_until_success
 
 class TestRouteReflectorCluster(TestBase):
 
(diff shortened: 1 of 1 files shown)


neiljerram:
@caseydavenport I don't yet understand the details here.  In the run that you linked, `test_route_reflector_cluster_resilience` failed, and if I'm reading the diags correctly

- it only allowed 7s for an inter-host ping after the last `create_bgp_peer`

- there were 6 attempts of `docker exec workload1 ping -c 1 -W 1 192.168.134.0`; so 1 original plus 5 retries.

But the code at that time was (and still is)

            # Peer the hosts with the route reflectors
            for host in [host1, host2]:
                for rr in rrc.get_redundancy_group():
                    create_bgp_peer(host, "node", rr.ip, 64513, metadata={'name': host.name + rr.name.lower()})

            # Assert network connectivity
            self.assert_true(workload_host1.check_can_ping(workload_host2.ip, retries=10))
            self.assert_ip_connectivity(workload_list=[workload_host1,
                                                       workload_host2],
                                        ip_pass_list=[workload_host1.ip,
                                                      workload_host2.ip])

So I don't understand why only 5 retries.

Secondly there's #114.  It was my bad there not to apply the same change to `test_route_reflector_cluster_resilience` - sorry - and I think we should do that.  Possibly then we wouldn't need _this_ PR as well; I don't know.  But if we do, it feels like we should wrap this 'wait until the BGP is really ready' stuff into common code, and use it consistently in all the relevant RR tests.  Otherwise I'm concerned that it will appear we had reasons (but unstated!) for doing slightly different things in different tests.

WDYT?

PR #129, opened 2018-12-19
Add foss checks make target

lwr20:
## Description
Adds make target to run foss license checks.

## Release Note
```release-note
None required
```


Makefile
```diff
@@ -326,6 +326,14 @@ static-checks: vendor
 fix:
 	goimports -w $(SRCFILES)
 
+foss-checks: vendor
```
fasaxc: Should this be called by the ci target?

lwr20: Good question.  I think it depends on whether we expect people to run `make ci` on their machines or if that target is only used on Semaphore.

If the `ci` target is only supposed to be used on Semaphore, then yes, I agree that I need to add `foss-checks` to the `ci` target.

@caseydavenport - care to make a call?

lwr20: I think the plan here is that we don't call this from the CI target and instead explicitly call from Semaphore on the master branch only.

PR #130, opened 2018-12-24
Update licenses folder with Fossa reports

lwr20:
## Description
Update licenses with reports from Fossa.

## Release Note

```release-note
None required
```


caseydavenport:
@lwr20 this looks fine to me, but it prompted me to think if the `calico/node` image is the right place for these files to live.

It feels like it might make more sense to compile them into the main Calico release archive, built as part of the final Calico release here: https://github.com/projectcalico/calico/blob/master/Makefile#L168

Also, can these be generated programatically?

lwr20:
Putting them in the release archive is fine by me, so long as that fulfils the terms of the licenses we're using.

> Also, can these be generated programatically?
Probably, but we haven't figured out how yet.  :)

PR #150, opened 2019-02-15
route filtering on rr testcase

roshanirathi:
This PR is to add a ST testcase to test the code changes done to resolve `RR filtering service advertisements` issue reported in confd repo.

### DIFF
--- a/tests/k8st/tests/test_bgp_advert.py
+++ b/tests/k8st/tests/test_bgp_advert.py
@@ -16,9 +16,11 @@ import os
 import subprocess
 import unittest
 from time import sleep
+from netaddr import IPAddress
 
 from kubernetes import client, config
 
+from tests.st.utils.docker_host import DockerHost
 from tests.k8st.test_base import TestBase
 from tests.k8st.utils.utils import start_external_node_with_bgp, \
         retry_until_success, run, curl, DiagsCollector
@@ -96,6 +98,56 @@ protocol bgp Mesh_10_192_0_4 from bgp_template {
 }
 """
 
+bird_conf2 = """
+router id 20.192.0.5;
+# Configure synchronization between routing tables and kernel.
+protocol kernel {
+  learn;             # Learn all alien routes from the kernel
+  persist;           # Don't remove routes on bird shutdown
+  scan time 2;       # Scan kernel routing table every 2 seconds
+  import all;
+  export all;
+  graceful restart;  # Turn on graceful restart to reduce potential flaps in
+                     # routes when reloading BIRD configuration.  With a full
+                     # automatic mesh, there is no way to prevent BGP from
+                     # flapping since multiple nodes update their BGP
+                     # configuration at the same time, GR is not guaranteed to
+                     # work correctly in this scenario.
+  merge paths on;
+}
+# Watch interface up/down events.
+protocol device {
+  debug { states };
+  scan time 2;    # Scan interfaces every 2 seconds
+}
+protocol direct {
+  debug { states };
+  interface -"cali*", "*"; # Exclude cali* but include everything else.
+}
+# Template for all BGP clients
+template bgp bgp_template {
+  debug { states };
+  description "Connection to BGP peer";
+  local as 64514;
+  multihop;
+  gateway recursive; # This should be the default, but just in case.
+  import all;        # Import all routes, since we don't know what the upstream
+                     # topology is and therefore have to trust the ToR/RR.
+  export all;
+  source address 20.192.0.5;  # The local address we use for the TCP connection
+  add paths on;
+  graceful restart;  # See comment in kernel section about graceful restart.
+  connect delay time 2;
+  connect retry time 5;
+  error wait time 5,30;
+}
+# ------------- Node-to-node mesh -------------
+# For peer /host/kube-master/ip_addr_v4
+protocol bgp Mesh_20_192_0_2 from bgp_template {
+  neighbor 20.192.0.2 as 64514;
+  passive on; # Mesh is unidirectional, peer will connect to us.
+}
+"""
 
 class TestBGPAdvert(TestBase):
 
@@ -273,3 +325,68 @@ EOF
                 for cip in cluster_ips:
                     self.assertNotIn(cip, routes)
             retry_until_success(check_routes_gone, retries=10, wait_time=5)
+
+    def test_rr_filtering_svc_adv(self):
+        """
+        Test route with svc ip is not filtered on RR
+        """
+        with DiagsCollector():
+            # Create a network policy on node-A accepts traffic from the external node.
+            run("""docker exec -i kube-master kubectl apply -f - << EOF
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: allow-tcp-80-ex
+  namespace: bgp-test
+spec:
+  podSelector: {node-A}
+  policyTypes:
+  - Ingress
+  ingress:
+  - from:
+    - ipBlock: { cidr: 20.192.0.5/32 }
+    ports:
+    - protocol: TCP
+      port: 80
+EOF
+""")
+           # Create RR host on node-B
+            cluster_id = str(IPAddress(0xFF000001))
+            rr = DockerHost('RR', start_calico=False)
+            rr.add_resource({
+                'apiVersion': 'projectcalico.org/v3',
+                'kind': 'Node',
+                'metadata': {
+                    'name': rr.get_hostname(),
+                    'labels': {
+                        'routeReflectorClusterID': cluster_id,
+                        },
+                    },
+                'spec': {
+                    'bgp': {
+                        'routeReflectorClusterID': cluster_id,
+                        },
+                    },
+                })
+            rr.start_calico_node()
+
+           # Disable node-to-node mesh
+            bgpconfig_as_num = 64514
+            peer_as_num = 64514
+            update_bgp_config(Node-A, nodeMesh=False, asNum=bgpconfig_as_num)
+            create_bgp_peer(Node-A, "global", rr.ip, peer_as_num)
+
+           # Allow peering from external node(node-C) to RR(node-B)
+            start_external_node_with_bgp("kube-node-extra", bird_conf2)
+
+           # Establish BGPPeer from cluster nodes to node-extra using calicoctl
+            run("""kubectl exec -i -n kube-system calicoctl -- /calicoctl apply -f - << EOF
+apiVersion: projectcalico.org/v3
+kind: BGPPeer
+metadata:
+  name: node-extra.peer
+spec:
+  peerIP: 20.192.0.5
+  asNumber: 64514
+EOF
+""")
--- a/tests/k8st/tests/test_bgp_advert.py
+++ b/tests/k8st/tests/test_bgp_advert.py
@@ -1,3 +1,4 @@
+#!/usr/bin/python
 # Copyright (c) 2018 Tigera, Inc. All rights reserved.
 #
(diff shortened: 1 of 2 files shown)


tests/k8st/tests/test_bgp_advert.py
```diff
@@ -273,3 +325,68 @@ def check_routes_gone():
...
+                        'routeReflectorClusterID': cluster_id,
+                        },
+                    },
+                'spec': {
+                    'bgp': {
+                        'routeReflectorClusterID': cluster_id,
+                        },
+                    },
+                })
+            rr.start_calico_node()
```
neiljerram: (As I've written elsewhere:) using DockerHost and rr.add_resource is wrong here because the compute nodes already exist in the k8st setup. So instead you need to get the Node resource of one of the existing nodes and modify it so that that node behaves as a RR. I experimented with code for some of that at https://github.com/projectcalico/node/pull/151, specifically

```
kubectl exec -i -n kube-system calicoctl -- /calicoctl get node kube-node-2 -o json
```

and you can see an example of what that does in the output for Job #2 at https://semaphoreci.com/calico/node/branches/pull-request-151/builds/3.

Building on that, you will still need to:
- get the JSON for the node into a Python dict form, instead of just printing it out into the log
- make the changes (to the dict) needed to configure RR behaviour
- apply the modified JSON back into Calico (with `calicoctl apply`)

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -273,3 +325,68 @@ def check_routes_gone():
...
+
+           # Disable node-to-node mesh
+            bgpconfig_as_num = 64514
+            peer_as_num = 64514
+            update_bgp_config(Node-A, nodeMesh=False, asNum=bgpconfig_as_num)
+            create_bgp_peer(Node-A, "global", rr.ip, peer_as_num)
```
neiljerram: These calls won't Just Work for the k8st test, because calicoctl needs to be called in a different way (as above).

You could just make different versions of these utilities, that call calicoctl in the right way.  But I wonder if we should instead look at modifying the k8st setup so that these calls do just work as they are.

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -273,3 +325,68 @@ def check_routes_gone():
...
+
+           # Allow peering from external node(node-C) to RR(node-B)
```
neiljerram: This comment is slightly.  In fact, the next line starts the external node and configures a BGP peering from it to the RR.  In order for that peering to succeed, it also needs to be added to the RR's own config.

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -273,3 +325,68 @@ def check_routes_gone():
...
+            start_external_node_with_bgp("kube-node-extra", bird_conf2)
+
+           # Establish BGPPeer from cluster nodes to node-extra using calicoctl
+            run("""kubectl exec -i -n kube-system calicoctl -- /calicoctl apply -f - << EOF
+apiVersion: projectcalico.org/v3
+kind: BGPPeer
+metadata:
+  name: node-extra.peer
+spec:
+  peerIP: 20.192.0.5
```
neiljerram: Is 20.192.0.5 the IP of the extra node?  Will it always be?

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -273,3 +325,68 @@ def check_routes_gone():
...
+  asNumber: 64514
+EOF
```
neiljerram: This BGPPeer will configure a peering from every Calico node to 20.192.0.5, whereas in fact we only want a peering from the RR.  If the RR is kube-node-2, the fix is to add
```
  node: kube-node-2
```
just before the peerIP line.

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -143,6 +228,112 @@ def assert_ecmp_routes(self, dst, via=["10.192.0.3", "10.192.0.4"]):
...
+""")
+
+        run("kubectl exec -i -n kube-system calicoctl -- /calicoctl get nodes -o yaml")
+        run("kubectl exec -i -n kube-system calicoctl -- /calicoctl get bgppeers -o yaml")
+        run("kubectl exec -i -n kube-system calicoctl -- /calicoctl get bgpconfigs -o yaml")
+
+        # Update the node-2 to behave as a route-reflector
+        run("kubectl exec -i -n kube-system calicoctl -- /calicoctl get node kube-node-2 -o json > /home/node.json")
+        run("sed -i '11 a \ \ \ \ \ \ \"i-am-a-route-reflector\": \"true\",' /home/node.json")
+        run("sed -i '21 a \ \ \ \ \ \ \"routeReflectorClusterID\": \"224.0.0.1\",' /home/node.json")
```
neiljerram: This works for now but is fragile in that it may break when new fields are added to the Node structure.  Instead I think we should:
- get the Node JSON as a string, from the `run` function:
  ```
  json_str = run("kubectl exec -i -n kube-system calicoctl -- /calicoctl get node kube-node-2 -o json")
  ```
- convert that to a Python dict:
  ```
  node_dict = json.loads(json_str)
  ```
- set the fields that we need for RR operation:
  ```
  node_dict['metadata']['labels']['i-am-a-route-reflector'] = 'true'
  node_dict['spec']['bgp']['routeReflectorClusterID'] = '224.0.0.1'
  ```
- convert back to a string, and apply:
```
run("""kubectl exec -i -n kube-system calicoctl -- /calicoctl apply -f - << 'EOF'
%s
EOF
""" % json.dumps(node_dict))
```

Do you agree that that would be more robust?

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -143,6 +228,112 @@ def assert_ecmp_routes(self, dst, via=["10.192.0.3", "10.192.0.4"]):
...
+- apiVersion: projectcalico.org/v3
+  kind: BGPConfiguration
+  metadata: {name: default}
+  spec:
+    nodeToNodeMeshEnabled: false
+    asNumber: 64512
+kind: BGPConfigurationList
+metadata:
+  resourceVersion: 139
+EOF
```
neiljerram: Please remove the List wrapper here, which isn't needed when writing a resource.  I.e. simplify to just this:
```
apiVersion: projectcalico.org/v3
kind: BGPConfiguration
metadata: {name: default}
spec:
  nodeToNodeMeshEnabled: false
  asNumber: 64512
```

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -143,6 +228,112 @@ def assert_ecmp_routes(self, dst, via=["10.192.0.3", "10.192.0.4"]):
...
+apiVersion: projectcalico.org/v3
+items:
+- apiVersion: projectcalico.org/v3
+  kind: BGPPeer
+  metadata: {name: kube-node-1}
+  spec:
+    node: kube-node-1
+    peerIP: 10.192.0.4
+    asNumber: 64512
+kind: BGPPeerList
```
neiljerram: Similarly here.

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -143,6 +228,112 @@ def assert_ecmp_routes(self, dst, via=["10.192.0.3", "10.192.0.4"]):
...
+EOF
+""")
+        run("kubectl get endpoints nginx-rr -n bgp-test -o yaml > /home/endpoint.yaml")
+        svcRoute = ""
+        with open("/home/endpoint.yaml", 'r') as stream:
+            content = yaml.load(stream)
+            for k,v in content.items():
+                if "clusterIP" in v:
+                    svcRoute = k
+        self.assertIn(svcRoute, self.get_routes())
```
neiljerram: I think we will need to allow some retries here, for the configuration and BGP updates to flow through.  So I think we need:
```
retry_until_success(lambda: self.assertIn(svcRoute, self.get_routes()))
```
That will try up to 10 times, with 1s interval.  If that isn't enough, we can use the `retries` and `wait_time` kwargs, for example:
```
retry_until_success(lambda: self.assertIn(svcRoute, self.get_routes()),
                                retries=12, wait_time=3)
```

roshanirathi: acknowledged

tests/k8st/tests/test_bgp_advert.py
roshanirathi: acknowledged

tests/k8st/tests/test_bgp_advert.py
roshanirathi: acknowledged

tests/k8st/tests/test_bgp_advert.py
roshanirathi: Thank you for suggesting this.

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -143,6 +228,112 @@ def assert_ecmp_routes(self, dst, via=["10.192.0.3", "10.192.0.4"]):
...
+      containers:
+      - name: nginx-rr
+        image: nginx:1.7.9
+        ports:
+        - containerPort: 80
+      nodeSelector:
+        beta.kubernetes.io/os: linux
+        kubernetes.io/hostname: kube-node-1
+---
+
```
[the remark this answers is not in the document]
roshanirathi: acknowledged

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -124,6 +182,33 @@ def setUp(self):
...
+        # Enable debug logging
+        self.update_ds_env("calico-node", "kube-system", "BGP_LOGSEVERITYSCREEN", "debug")
+
+        # Establish BGPPeer from cluster nodes to node-extra using calicoctl
+        run("""kubectl exec -i -n kube-system calicoctl -- /calicoctl apply -f - << EOF
+apiVersion: projectcalico.org/v3
+kind: BGPPeer
+metadata:
+  name: node-extra.peer
+spec:
```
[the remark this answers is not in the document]
roshanirathi: acknowledged

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -124,6 +182,33 @@ def setUp(self):
...
+
+        start_external_node_with_bgp("kube-node-extra", bird_conf_rr)
+
+        # set CALICO_ADVERTISE_CLUSTER_IPS=10.96.0.0/12
+        self.update_ds_env("calico-node", "kube-system", "CALICO_ADVERTISE_CLUSTER_IPS", "10.96.0.0/12")
+
+        # Enable debug logging
+        self.update_ds_env("calico-node", "kube-system", "BGP_LOGSEVERITYSCREEN", "debug")
+
+        # Establish BGPPeer from cluster nodes to node-extra using calicoctl
```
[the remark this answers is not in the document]
roshanirathi: The bird_conf_rr that is used to create the external-node has source ip as 10.192.0.5. So it will get the same ip every time.

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -96,6 +100,60 @@
 }
 """
```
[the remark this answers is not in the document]
roshanirathi: acknowledged

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/python
```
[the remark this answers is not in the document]
roshanirathi: acknowledged

PR #152, opened 2019-02-18
Rationalize ST calicoctl calling code

neiljerram:
Just deleting some duplicate code here.  But with an eye on moving towards a clearer abstraction of calicoctl operations, so that we might eventually be able to use useful functions like update_bgp_config in both the tests/st and tests/k8st suites.  (Currently we can't do that, because the code underneath update_bgp_config boils down to
```
docker exec <host container> calicoctl ...
```
whereas the setup for tests/k8st requires
```
kubectl exec -i -n kube-system calicoctl -- /calicoctl ...
```
)


neiljerram:
@caseydavenport Sorry, yes, I just lost track of this.  Will merge now.

PR #154, opened 2019-02-21
Assign attributes to IPIP address

caseydavenport:
## Description

Requires this PR first: https://github.com/projectcalico/libcalico-go/pull/1035

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/allocateipip/allocate_ipip_addr.go
+++ b/pkg/allocateipip/allocate_ipip_addr.go
@@ -2,6 +2,7 @@ package allocateipip
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"time"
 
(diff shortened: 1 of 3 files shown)


pkg/allocateipip/allocate_ipip_addr.go
```diff
@@ -155,11 +155,15 @@ func removeHostTunnelAddr(ctx context.Context, c client.Interface, nodename stri
...
+	attrs := map[string]string{
+		ipam.AttributeNode: nodename,
+		ipam.AttributeType: ipam.AttributeTypeIPIP,
+	}
 	args := ipam.AutoAssignArgs{
 		Num4:      1,
 		Num6:      0,
-		HandleID:  nil,
-		Attrs:     nil,
+		HandleID:  &nodename,
```
caseydavenport: Probably a better handle than just the nodename, but couldn't think of one in a few seconds.

fasaxc: Suggest sticking some prefix on it.  Might come in useful later; will certainly make it easier to see what that handle was for in a dump.  `ipip-tunnel-`?

PR #162, opened 2019-03-04
Added code to avoid aggresive update due to rolling update issue on any node.

tathagatachowdhury:
This is a bug fix to check BGP readiness and avoid aggressive update if during rolling update there is any issue to update a particular node.

### DIFF
--- a/cmd/calico-node/main.go
+++ b/cmd/calico-node/main.go
@@ -46,7 +46,7 @@ var bird6Ready = flagSet.Bool("bird6-ready", false, "Run BIRD6 readiness checks"
 var felixReady = flagSet.Bool("felix-ready", false, "Run felix readiness checks")
 
 // thresholdTime is introduced for bird readiness check. Default value is 30 sec.
-var thresholdTime = flagSet.Duration("thresholdTime", 30*time.Second, "Threshold time for bird readiness")
+var thresholdTime = flagSet.Duration("threshold-time", 30*time.Second, "Threshold time for bird readiness")
 
 // confd flags
 var runConfd = flagSet.Bool("confd", false, "Run confd")
--- a/cmd/calico-node/main.go
+++ b/cmd/calico-node/main.go
@@ -17,6 +17,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"time"
 
 	confdConfig "github.com/kelseyhightower/confd/pkg/config"
 	confd "github.com/kelseyhightower/confd/pkg/run"
@@ -44,6 +45,9 @@ var birdReady = flagSet.Bool("bird-ready", false, "Run BIRD readiness checks")
 var bird6Ready = flagSet.Bool("bird6-ready", false, "Run BIRD6 readiness checks")
 var felixReady = flagSet.Bool("felix-ready", false, "Run felix readiness checks")
 
+// thresholdTime is introduced for bird readiness check. Default value is 30 sec.
+var thresholdTime = flagSet.Duration("thresholdTime", 30*time.Second, "Threshold time for bird readiness")
+
 // confd flags
 var runConfd = flagSet.Bool("confd", false, "Run confd")
 var confdRunOnce = flagSet.Bool("confd-run-once", false, "Run confd in oneshot mode")
@@ -63,6 +67,9 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Threshold time for bgp peering
+	fmt.Println("Threshold time for bird readiness check: ", *thresholdTime)
+
 	// Perform some validation on the parsed flags. Only one of the following may be
 	// specified at a time.
 	onlyOne := []*bool{version, runFelix, runStartup, runConfd}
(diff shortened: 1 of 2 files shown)


pkg/readiness/readiness.go
```diff
@@ -60,30 +59,48 @@ func Run(bird, bird6, felix bool) {
...
 
+	// numEstablishedPeer keeps count of number of peers with bgp state established
+	numEstablishedPeer := 0
 	s := []string{}
+
 	for _, peer := range peers {
-		if peer.BGPState != "Established" {
+		if peer.BGPState == "Established" {
+			numEstablishedPeer += 1
+		} else if peer.BGPState != "Established" {
```
fasaxc: The `if peer.BGPState != "Established"` is redundant here since the only way to reach the `else` block is if `peer.BGPState != "Established"`

pkg/readiness/readiness.go
```diff
@@ -60,30 +59,48 @@ func Run(bird, bird6, felix bool) {
...
 			s = append(s, peer.PeerIP)
 		}
 	}
-	if len(s) > 0 {
-		return fmt.Errorf("BGP not established with %+v", strings.Join(s, ","))
-	}
 
-	// Check for GR
-	gr, err := bird.GRInProgress(ipv)
+	fmt.Printf("Number of nodes with BGP peering established = %v", numEstablishedPeer)
```
fasaxc: Should use `log.Debugf` or `log.Infof`, I think; that seems to be consistent with what the rest of this code does.

pkg/readiness/readiness.go
```diff
@@ -60,30 +59,48 @@ func Run(bird, bird6, felix bool) {
...
+
+	nodenameFileStat, err := os.Stat("/var/lib/calico/nodename")
 	if err != nil {
-		return err
-	} else if gr {
-		return errors.New("graceful restart in progress")
+		return fmt.Errorf("Error in nodename file: %v", err)
```
fasaxc: How about "Failed to stat nodename file"

pkg/readiness/readiness.go
```diff
@@ -60,30 +59,48 @@ func Run(bird, bird6, felix bool) {
...
+	}
+
+	if time.Since(nodenameFileStat.ModTime()) < thresholdTime {
+		if len(s) > 0 {
```
fasaxc: Think this needs a comment to explain why we're doing a different check here.
```
// When we first start up, only report ready if all our peerings are established.
// This prevents rolling update from proceeding until BGP is back up.
```

pkg/readiness/readiness.go
```diff
@@ -60,30 +59,48 @@ func Run(bird, bird6, felix bool) {
...
+		}
+		// Check for GR
+		gr, err := bird.GRInProgress(ipv)
+		if err != nil {
+			return err
+		} else if gr {
+			return errors.New("graceful restart in progress")
+		}
+	} else if numEstablishedPeer > 0 {
+		log.Debugf("There exist(s) %v calico node which has BGP peering established.", numEstablishedPeer)
```
fasaxc: Same here:
```
// After a while, only require a single peering to be up.  This prevents the whole mesh
// from reporting not-ready if some nodes go down.
```

cmd/calico-node/main.go
```diff
@@ -44,6 +45,9 @@ var birdReady = flagSet.Bool("bird-ready", false, "Run BIRD readiness checks")
 var bird6Ready = flagSet.Bool("bird6-ready", false, "Run BIRD6 readiness checks")
 var felixReady = flagSet.Bool("felix-ready", false, "Run felix readiness checks")
 
+// thresholdTime is introduced for bird readiness check. Default value is 30 sec.
+var thresholdTime = flagSet.Duration("thresholdTime", 30*time.Second, "Threshold time for bird readiness")
```
fasaxc: For consistency with other params above, `"thresholdTime"` should be `"threshhold-time"`

pkg/readiness/readiness.go
```diff
@@ -60,30 +59,48 @@ func Run(bird, bird6, felix bool) {
...
-func checkBIRDReady(ipv string) error {
+func checkBIRDReady(ipv string, thresholdTime time.Duration) error {
 	// Check for unestablished peers
 	peers, err := bird.GetPeers(ipv)
 	log.Debugf("peers: %v", peers)
 	if err != nil {
 		return err
 	}
 
+	// numEstablishedPeer keeps count of number of peers with bgp state established
```
fasaxc: How about moving the Stat() up to this point itn he function, then you can move the rest of the function to a new utility function, which can be easily UTed.

I.e. the new utility function should take arguments for the BGP peers and the node startup time.

cmd/calico-node/main.go
tathagatachowdhury: Acknowledged and updated.

pkg/readiness/readiness.go
tathagatachowdhury: Acknowledged and updated.

pkg/readiness/readiness.go
tathagatachowdhury: Acknowledged and updated.

pkg/readiness/readiness.go
tathagatachowdhury: Acknowledged and added.

pkg/readiness/readiness.go
tathagatachowdhury: Acknowledged and added.

PR #163, opened 2019-03-04
Have all k8st resources locally and update them so that CI runs correctly

lwr20:
## Description
Now that the hosted etcd manifest is no longer in master, we need replacement resources locally in order to run CI.

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -481,7 +481,7 @@ k8s-test:
 ## Start k8s cluster
 k8s-start: $(NODE_CONTAINER_CREATED) tests/k8st/$(DIND_SCR)
 	CNI_PLUGIN=calico \
-	CALICO_VERSION=master \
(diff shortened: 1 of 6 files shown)


caseydavenport:
@lwr20 once merged can you also revert https://github.com/projectcalico/calico/pull/2465?

lwr20:
> Longer term, I would much prefer to not check in so much of this, specifically the dind-cluster script
The difficulties here:

1. upstream are VERY slow to merge contributions (Neil's PR has been open for ~6 months)
2. the checked in script is NOT the upstream one - I had to modify it to work with local yaml files (upstream script has hard-coded the CNI manifest URLs).  To upstream that, we'd need to submit a PR putting all the manifest URLs into overridable env vars (and see point 1 above)

PR #168, opened 2019-03-06
[V3.3]Backporting changes from master for bird readiness check to avoid aggresive update.

tathagatachowdhury:
This is a back-porting of changes from master for bird readiness check to avoid aggressive update in case of issue during rolling update.


caseydavenport:
https://github.com/projectcalico/node/pull/162

PR #207, opened 2019-04-06
Fix for ipset error

stevegaossou:
## Description
Fixes issue with ipset when running add commands that include tcp, to avoid error "Syntax error: 'tcp' is invalid as number".

### Steps to Reproduce
- On `master` build the node image 
```
make image
```
- Now run it in privileged mode
```
docker run -it --privileged calico/node:latest-amd64 sh
```
- Run the following: 
```
# ipset create cali4t47 hash:ip,port family inet maxelem 1048576
# ipset add cali4t47 192.168.85.14,tcp:80
```
- Observe the error: 
```
ipset v6.38: Syntax error: 'tcp' is invalid as number
Syntax error: cannot parse 'tcp' as a protocol
```

### Fix
- In the same container install 
```
# apt-get install netbase
```
- Now run the second command again (it should now work without the error)
```
# ipset add cali4t47 192.168.85.14,tcp:80
#
```
## Todos
- [ ] ~~Tests~~
- [ ] ~~Documentation~~
- [ ] ~~Release note~~

## Release Note
```release-note
None required
```

### DIFF
--- a/Dockerfile.amd64
+++ b/Dockerfile.amd64
@@ -34,6 +34,9 @@ RUN apt-get update && apt-get install -y \
     runit \ 
     # Need kmod to ensure ip6tables-save works correctly
     kmod \ 
+    # Need netbase in order for ipset to work correctly 
+    # See https://github.com/kubernetes/kubernetes/issues/68703
+    netbase \ 
     # Also needed (provides utilities for browsing procfs like ps) 
     procps \    
     ca-certificates


stevegaossou:
Hey @caseydavenport I don't have permissions so I cannot remove the labels `docs-pr-required` and `release-not-required`. I'm guessing I don't need those for this fix. 

Also, as we discussed this probably doesn't require adding a regression test, since we already have some in Wavetank.

PR #208, opened 2019-04-11
Add new tool to the image and bump Felix

iaguis:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -32,10 +32,12 @@ import:
   - lib/options
   - lib/upgrade/migrator
   - lib/upgrade/migrator/clients
-# gobuffalo/logger needs the SetOutput function in the Logger type, so pin
-# the first version that introduces it.
+# Use our fork, because v1.0.5 causes test breakages due to problems with our log formatting code,
+# but other deps require newer version because of the SetOutput function. Our fork backports
+# the missing stuff on top of v1.0.4.
 - package: github.com/sirupsen/logrus
-  version: v1.0.6
+  repo: https://github.com/projectcalico/logrus
+  version: v1.0.4-calico
 - package: k8s.io/apimachinery
   subpackages:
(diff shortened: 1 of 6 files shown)


glide.yaml
```diff
@@ -32,8 +32,10 @@ import:
   - lib/options
   - lib/upgrade/migrator
   - lib/upgrade/migrator/clients
+# gobuffalo/logger needs the SetOutput function in the Logger type, so pin
+# the first version that introduces it.
 - package: github.com/sirupsen/logrus
-  version: v1.0.4
+  version: v1.0.6
```
caseydavenport: @fasaxc didn't we pin to v1.0.4 specifically since newer versions break some of our logging code?

Do we need to do more here before we can rev this?

iaguis: From OOB conversation with @fasaxc, the issue (https://github.com/sirupsen/logrus/issues/729) seems to be fixed in 1.0.6 (https://github.com/sirupsen/logrus/commit/070c81def33f6362a8267b6a4e56fb7bf23fc6b5) so we should be fine here.

caseydavenport: Hm, I'm not seeing the fix work, though I might be doing something wrong. Here's some output from a Felix FV log using the new logrus:

```
[INFO][11] <nil> <nil>: Parsing value for LogSeverityFile: None (from config file)
```

fasaxc: Sigh

krnowak: All the logging using formatting versions of the functions (`Debugf`, `Infof`, `Warnf`) ended up having `<nil> <nil>` for some reason. Using `Debug` or `Info` works just fine. So I suppose the issue is somewhere else in logrus.

Fix one thing, break another. :)

PR #209, opened 2019-04-11
Add a "vxlan" (only) networking mode.

fasaxc:
## Description

In vxlan mode, we still allocate the (now misnamed) BGP node IP but we don't run BIRD.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


caseydavenport:
Is there a corresponding docs PR for this as well?

PR #219, opened 2019-04-25
add bs4 to calico/test image

2ffs2nns:
Need to add bs4 for some additional tests


caseydavenport:
I think we said we won't add this here, since it's for tests in another repo.

PR #226, opened 2019-05-06
Use specific branch of calicoctl

caseydavenport:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -113,8 +113,8 @@ BIRD_VER?=v0.3.3-0-g1e8dd375
 BIRD_IMAGE ?= calico/bird:$(BIRD_VER)-$(ARCH)
 
 # Versions and locations of dependencies used in tests.
-CALICOCTL_VER?=master
-CNI_VER?=master
+CALICOCTL_VER?=$(shell git rev-parse --abbrev-ref HEAD)
+CNI_VER?=$(shell git rev-parse --abbrev-ref HEAD)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -113,8 +113,8 @@ BIRD_VER?=v0.3.3-0-g1e8dd375
 BIRD_IMAGE ?= calico/bird:$(BIRD_VER)-$(ARCH)
 
 # Versions and locations of dependencies used in tests.
-CALICOCTL_VER?=master
-CNI_VER?=master
+CALICOCTL_VER?=$(shell git rev-parse --abbrev-ref HEAD)
```
tomastigera: what if I am not in a branch with a common name across repos like release-3.x but in my own my-bug-fix branch when executing make st? Its going to pull a calicoctl image that does not exist?

caseydavenport: Yeah it is, which kinda stinks.

We could maybe check if it's something non-standard and revert back to master?

or, we just manually pin this when cutting release branches... it's all icky.

FWIW we do this same thing for the update-libcalico target, so there is some precedent.

tomastigera: Yeah, it works for releases, but breaks regular dev-workflow. Setting CTL_CONTAINER_NAME would work for releases too while keeping the default master for regular development ...

PR #231, opened 2019-05-06
Fix calicoctl update and usage

tomastigera:
- clean must remove dist/calicoctl to force updating it

- calicoctl version reports verion used properly

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -167,6 +167,7 @@ clean:
 	find . -name '*.created' -exec rm -f {} +
 	find . -name '*.pyc' -exec rm -f {} +
 	rm -rf certs *.tar vendor $(NODE_CONTAINER_BIN_DIR)
+	rm -rf dist
 
 	# Delete images that we built in this repo
 	docker rmi $(BUILD_IMAGE):latest-$(ARCH) || true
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -160,6 +160,7 @@ clean:
 	find . -name '*.created' -exec rm -f {} +
 	find . -name '*.pyc' -exec rm -f {} +
 	rm -rf certs *.tar vendor $(NODE_CONTAINER_BIN_DIR)
+	rm dist/calicoctl
```
jpalardy:
```suggestion
	rm -f dist/calicoctl
```

without `-f`, the exit code of rm != 0 if the file is missing.

Should the whole `dist` directory be deleted?

tomastigera: - or -f ... ack

I would guess dist/*, waiting for opinion from @caseydavenport & co.

caseydavenport: Yeah I think we can blow away all of `dist` in `clean`

tomastigera: ack, done

PR #235, opened 2019-05-08
Check for OOM kills when a test fails

neiljerram:


neiljerram:
@caseydavenport Would you like to review and maybe merge this? If the prominent remaining flakes are caused by OOM, I'm hopeful that this will make that clear.

PR #241, opened 2019-05-28
Configurable port for Felix liveness / readiness probes

jadech32:
## Description

- Changed liveness probe from http request to command
- Health Port now defined as environment variable `FELIX_HEALTHPORT` in `calico`
- Liveness and readiness probes now look at either default value or environment variable (if set correctly)

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix Felix readiness checks when using non-default port
```


pkg/liveness/liveness.go
```diff
@@ -0,0 +1,58 @@
...
+	"net/http"
+	"os"
+	"strconv"
+	"time"
+)
+
+var felixLivenessEp string
+
+func init() {
+	felixPort := os.Getenv("FELIX_HEALTHPORT")
```
jpalardy: maybe this would be better as `os.LookupEnv` to distinguish from an explicit blank `FELIX_HEALTHPORT`?

pkg/liveness/liveness.go
```diff
@@ -0,0 +1,58 @@
...
+
+// CheckFelixLive Checks whether or not felix is live.
+func CheckFelixLive() (err error) {
+	c := &http.Client{Timeout: 5 * time.Second}
+	resp, err := c.Get(felixLivenessEp)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if e := resp.Body.Close(); e != nil {
```
jpalardy: I wonder if this whole defer block can be reduced to:

```go
defer func() {
     err = resp.Body.Close()
}
```

pkg/readiness/readiness.go
```diff
@@ -26,7 +27,21 @@ import (
 	"github.com/projectcalico/node/pkg/readiness/bird"
 )
 
-const felixReadinessEp = "http://localhost:9099/readiness"
+var felixReadinessEp string
+
+func init() {
+	felixPort := os.Getenv("FELIX_HEALTHPORT")
```
caseydavenport: I think we need to support more than just the environment variable, since Felix configuration parameters can also be configured via config file or datastore config. There might be some Felix code we can import / copy here to make sure we're using the same order of operations?

CC @fasaxc

pkg/liveness/liveness.go
fasaxc: I don't think it can.  In the case that `Close()` returns `nil` but the status code is bad, I think we'd hit this line:
```
return fmt.Errorf("liveness probe reporting %d", resp.StatusCode)
```
and then overwrite `err` with `nil` when the defer runs, suppressing the error.

pkg/readiness/readiness.go
fasaxc: Might be more trouble than its worth.  Env vars override the other config mechanisms so every env var in our manifest prevents other config mechanisms from having an effect.  The manifest seems like a sensible place to co-configure felix and the liveness checker.

pkg/liveness/liveness.go
```diff
@@ -0,0 +1,58 @@
+// Copyright (c) 2018 Tigera, Inc. All rights reserved.
```
fasaxc: Should be 2019

pkg/liveness/liveness.go
fasaxc: Good thing to think about but I can't see any reason to distinguish the two cases here; go idiom is to make the zero value useful. In this case no env var and env var equal to `""` can both mean "use the default"; WDYT?

pkg/liveness/liveness.go
jpalardy: Right... I over-optimized the defer.

pkg/liveness/liveness.go
jpalardy: I was trying to imagine what a user would expect from `FELIX_HEALTHPORT="" node-calico [...]`

_(in practice, it would probably be the result of some other bad config.)_

Flagging empty when the `Atoi` fails just below might be friendlier than falling back on the default.

But I'm not against the blank field defaulting :smile:

jadech32: I think `os.Getenv` and `os.LookupEnv` would result in the same logic here since for our case we use the default for all cases except where the environment variable is set correctly, so I don't see a reason to use `os.LookupEnv` over `os.Getenv`

pkg/liveness/liveness.go
```diff
@@ -0,0 +1,58 @@
...
+
+func init() {
+	felixPort := os.Getenv("FELIX_HEALTHPORT")
+	if felixPort == "" {
+		felixLivenessEp = "http://localhost:9099/liveness"
+		return
+	}
+
+	if _, err := strconv.Atoi(felixPort); err != nil {
+		log.Panicf("Failed to parse value for port %s", felixPort)
```
jpalardy: That would probably be clearer as `%q`

pkg/readiness/readiness.go
```diff
@@ -26,7 +27,21 @@ import (
...
+	if felixPort == "" {
+		felixReadinessEp = "http://localhost:9099/readiness"
+		return
+	}
+
+	if _, err := strconv.Atoi(felixPort); err != nil {
+		log.Panicf("Failed to parse value for port %s", felixPort)
```
jpalardy: That would probably be clearer as `%q`

jadech32:
> Code looks correct at the small scale, but I think there's a problem when it's put together with the readiness check: as soon as you set the new env var, the readiness check will fail because it has a [hardcoded port](https://github.com/projectcalico/node/blob/master/pkg/readiness/readiness.go#L29) (and the liveness and readiness endpoints share a port).
> 
> You could just address that issue by altering the readiness checking code but it'd be cleaner to refactor so that the readiness and liveness checking logic is shared (algorithm should be exactly the same, apart from the URL path that they query and the logs they emit).

Regarding the hardcoded port issue, I've already made changes to grab the port from the environment variable. As for the refactoring, I was thinking about it but i was hesitant since the existing code was under the `readiness` directory (`readiness.go`)

fasaxc:
Good point, why not rename the directory `health` and parametrise the checker function on `liveness` vs `readiness`?

pkg/readiness/readiness.go
caseydavenport: Yeah, could be. Just feels like something that will come back to bite us. I'd prefer not to leave that bit of debt lying around.

cmd/calico-node/main.go
```diff
@@ -82,9 +85,15 @@ func main() {
...
+	// Check for liveness
+	if *felixLive {
+		health.Run(*birdReady, *bird6Ready, *felixReady, *felixLive, *thresholdTime)
+		os.Exit(0)
+	}
+
 	// If any of the readienss options are provided, check readiness.
 	if *birdReady || *bird6Ready || *felixReady {
-		readiness.Run(*birdReady, *bird6Ready, *felixReady, *thresholdTime)
+		health.Run(*birdReady, *bird6Ready, *felixReady, false, *thresholdTime)
```
fasaxc: Seems a bit odd to have two blocks here when we already have the relevant bools in hand.  Maybe just do:
```
if *felixLive || *birdReady || *bird6Ready || *felixReady { 
    health.Run(*birdReady, *bird6Ready, *felixReady, *felixLive, *thresholdTime)
    os.Exit(0)
}
```

jadech32: Fixed.

jiuchen1986:
Hi. I just wonder why we change the liveness probe from httpGet to exec command here. Any background explanations here? Many thanks. @jadech32

fasaxc:
@jiuchen1986 it's so we can check the liveness of multiple components inside calico/node.  You can only have one HTTP check but you can have an exec check that then does more than HTTP (or other) check.

PR #263, opened 2019-07-02
Add spoof tests

lwr20:
## Description
Add tests to ensure spoofing encapsulated packets is not possible

## Todos
- [x] Tests

## Release Note
```release-note
None required
```

### DIFF
--- a/tests/k8st/tests/test_ipip_spoofing.py
+++ b/tests/k8st/tests/test_ipip_spoofing.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2018 Tigera, Inc. All rights reserved.
+# Copyright (c) 2019 Tigera, Inc. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -33,7 +33,7 @@ class ConnectionError(Exception):
     pass
 
 
-class TestIPIPSpoof(TestBase):
+class TestSpoof(TestBase):
     def setUp(self):
         TestBase.setUp(self)
         self.ns_name = generate_unique_id(5, prefix="spoof")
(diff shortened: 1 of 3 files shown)


tests/k8st/tests/test_ipip_spoofing.py
```diff
@@ -0,0 +1,229 @@
...
+    retry_until_success, \
+    DiagsCollector, \
+    calicoctl, \
+    calicoctl_apply_dict, \
+    kubectl, \
+    generate_unique_id
+
+_log = logging.getLogger(__name__)
+
+
```
rafaelvanoni: Couple extra blank lines here

tests/k8st/tests/test_ipip_spoofing.py
```diff
@@ -0,0 +1,229 @@
...
+        # restart calico-nodes
+        kubectl("delete po -n kube-system -l k8s-app=calico-node")
+        kubectl("wait --timeout=2m --for=condition=ready" +
+                " pods -l k8s-app=calico-node -n kube-system")
+
+    def test_simple_ipip_spoof(self):
+        with DiagsCollector():
+            # Change pool to use IPIP
+            default_pool = json.loads(calicoctl("get ippool default-ipv4-ippool -o json"))
+            default_pool["spec"]["vxlanMode"] = "Never"
```
rafaelvanoni: Just a nit, but since IPIP is the default it might worth checking if we need to apply this first.

tests/k8st/tests/test_ipip_spoofing.py
lwr20: My PEP-8 linter (pycharm) says these blank lines are necessary:
https://www.python.org/dev/peps/pep-0008/#blank-lines says:
> Surround top-level function and class definitions with two blank lines.

PR #264, opened 2019-07-02
Use iptables 1.8 with amd64 image

tmjd:
## Description
Update amd64 image to use iptables 1.8

## Release Note

```release-note
Use iptables 1.8 with amd64.
```

### DIFF
--- a/Dockerfile.amd64
+++ b/Dockerfile.amd64
@@ -21,12 +21,6 @@ LABEL maintainer "Casey Davenport <casey@tigera.io>"
 
 ARG ARCH=amd64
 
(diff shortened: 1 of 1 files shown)


Dockerfile.amd64
```diff
@@ -46,6 +40,16 @@ RUN apt-get update && apt-get install -y \
     procps \    
     ca-certificates
 
+RUN echo 'APT::Default-Release "stable";' > /etc/apt/apt.conf.d/99defaultrelease
```
fasaxc: Installing a backport is a bit unusual so please add a comment explaining it (as we had for the 1.6.2 backport)

PR #283, opened 2019-08-06
Enable VXLAN when VXLANMode is CrossSubnet

Brian-McM:
This should mimic the behaviour of when VXLANMode is Always

## Description
We would like to have a CrossSubnet option available for VXLAN that works much the same way the the CrossSubnet option works for IPIP. This PR tells node to recognise the CrossSubnet for VXLAN, and do what it does for the Always mode.
-->

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Might need release notes
```

### DIFF
--- a/pkg/allocateip/allocateip_test.go
+++ b/pkg/allocateip/allocateip_test.go
@@ -435,6 +435,36 @@ var _ = Describe("determineEnabledPoolCIDRs", func() {
 			Expect(cidrs).To(ContainElement(*cidr1))
 			Expect(cidrs).ToNot(ContainElement(*cidr2))
 		})
+		It("should match ip-pool-1 but not ip-pool-2 for VXLANMode CrossSubnet", func() {
+			// Mock out the node and ip pools
+			n := api.Node{ObjectMeta: metav1.ObjectMeta{Name: "bee-node", Labels: map[string]string{"foo": "bar"}}}
+			pl := api.IPPoolList{
+				Items: []api.IPPool{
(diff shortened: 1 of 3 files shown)


pkg/allocateip/allocateip_test.go
```diff
@@ -435,6 +435,36 @@ var _ = Describe("determineEnabledPoolCIDRs", func() {
...
+						Spec: api.IPPoolSpec{
+							Disabled:     false,
+							CIDR:         "172.128.0.0/9",
+							NodeSelector: `foo != "bar"`,
+							VXLANMode:    api.VXLANModeCrossSubnet,
+						},
+					}}}
+
+			// Execute and test assertions.
+			cidrs := determineEnabledPoolCIDRs(n, pl, true)
```
caseydavenport: I don't understand this test fully. I think it's testing the `NodeSelector` field rather than the `VXLANModeCrossSubnet` field, IIUC.

I think we want two pools which select this node - one with no VXLAN and one with CrossSubnet, and then assert that the CrossSubnet one is selected?

Brian-McM: The test is to make sure that a VXLAN pool with VXLANMode CrossSubnet is selected in the same way the Always is. So this tests that a matching CrossSubnet pool is selected, and that a none matching CrossSubnet pool is not (just to make sure the does not match logic still works for CrossSubnet)

PR #291, opened 2019-08-15
Update OSS semaphore jobs that depend on glide

rafaelvanoni:


Makefile
```diff
@@ -176,8 +176,16 @@ SRC_FILES=$(shell find ./pkg -name '*.go')
 
 EXTRA_DOCKER_ARGS	+= -e GO111MODULE=on
 
-ifdef GOPATH
-	EXTRA_DOCKER_ARGS += -v $(GOPATH)/pkg/mod:/go/pkg/mod:rw
+# Volume-mount gopath into the build container if it's explicitly set for persistent caching.
```
caseydavenport: Do we use circle CI in node? I didn't think we did.

rafaelvanoni: I'm adding the same checks everywhere, I'm assuming we don't have a set strategy for which repos will eventually have CircleCI jobs or not.

PR #306, opened 2019-08-28
Run 'go mod download' after updating pins to workaround issue with co…

rafaelvanoni:


caseydavenport:
@rafaelvanoni I think we might still need this or something like it to make the update-pins target work. 

Looks like this needs a rebase though.

PR #323, opened 2019-09-10
K8ST: Better code for DaemonSet update

neiljerram:
This PR improves the K8ST update_ds_env function so that it can change
more than one environment variable, in the calico-node DaemonSet, at
once.  We don't strictly need it now, as we have stopped setting the
CALICO_ADVERTISE_CLUSTER_IPS variable, but I believe it's a useful
capability and would be a shame to throw away.

Following is my previous commit message from when I worked on this in
September 2010:

We've been seeing flakiness in this area in Semaphore test runs.
Specifically:

- when `TestBGPAdvert.test_mainline` runs on Semaphore, it appears to
hang for a long time here, and sometimes the whole test run is timed
out

- when I've enabled it, logging shows that the DaemonSet updates seem
to take a long time, with logging like this:

```
    2019-09-09 10:29:34,636 INFO [2019-09-09 10:29:34.636901] docker exec kube-node-extra service bird restart
    2019-09-09 10:29:34,729 INFO Output:

    2019-09-09 10:29:44,765 INFO 3/3 nodes updated
    2019-09-09 10:29:54,808 INFO 1/3 nodes updated
    2019-09-09 10:30:04,826 INFO 1/3 nodes updated
    2019-09-09 10:30:14,845 INFO 1/3 nodes updated
    2019-09-09 10:30:24,862 INFO 1/3 nodes updated
    2019-09-09 10:30:34,881 INFO 1/3 nodes updated
    2019-09-09 10:30:44,899 INFO 1/3 nodes updated
    2019-09-09 10:30:54,917 INFO 1/3 nodes updated
    2019-09-09 10:31:04,935 INFO 1/3 nodes updated
    2019-09-09 10:31:14,961 INFO 1/3 nodes updated
    2019-09-09 10:31:24,986 INFO 1/3 nodes updated
    ...
```

I wondered if it could be problematic to make two updates so quickly
to the same DaemonSet.  Anyway, it feels lazy, so I looked at changing
the code to set both of the variables that we need in a single
update.  The obvious change here unexpectedly raised an error, and I
eventually concluded that

    container.env.append({"name": k, "value": v, "value_from": None})

was wrong, and should be

    v1_ev = client.V1EnvVar(name=k, value=v, value_from=None)
    container.env.append(v1_ev)

I haven't worked out for sure if the previous form ever worked, and if
so why.  (Perhaps there is some special casing to allow the plain dict
as well as a V1EnvVar?  Perhaps one of the Python packages that we use
has been revved?)

Anyway, the new code is definitely better, by correctly using a
V1EnvVar, and by setting both variables in a single DaemonSet update.
Let's see if it helps with the observed flakiness.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/tests/k8st/test_base.py
+++ b/tests/k8st/test_base.py
@@ -208,18 +208,21 @@ class TestBase(TestCase):
     def get_routes(self):
         return run("docker exec kube-node-extra ip r")
 
-    def update_ds_env(self, ds, ns, k, v):
+    def update_ds_env(self, ds, ns, env_vars):
         config.load_kube_config(os.environ.get('KUBECONFIG'))
         api = client.AppsV1Api(client.ApiClient())
         node_ds = api.read_namespaced_daemon_set(ds, ns, exact=True, export=True)
         for container in node_ds.spec.template.spec.containers:
             if container.name == ds:
-                env_present = False
-                for env in container.env:
-                    if env.name == k:
-                        env_present = True
-                if not env_present:
-                    container.env.append({"name": k, "value": v, "value_from": None})
+                for k, v in env_vars.items():
+                    logger.info("Set %s=%s", k, v)
+                    env_present = False
+                    for env in container.env:
+                        if env.name == k:
+                            env_present = True
+                    if not env_present:
+                        v1_ev = client.V1EnvVar(name=k, value=v, value_from=None)
+                        container.env.append(v1_ev)
         api.replace_namespaced_daemon_set(ds, ns, node_ds)
 
         # Wait until the DaemonSet reports that all nodes have been updated.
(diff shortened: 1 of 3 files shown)


tests/k8st/test_base.py
```diff
@@ -204,18 +204,21 @@ class StillThere(Exception):
...
-                    container.env.append({"name": k, "value": v, "value_from": None})
+                for k, v in env_vars.items():
+                    logger.info("Set %s=%s", k, v)
+                    env_present = False
+                    for env in container.env:
+                        if env.name == k:
+                            env_present = True
+                    if not env_present:
+                        v1_ev = client.V1EnvVar(name=k, value=v, value_from=None)
+                        container.env.append(v1_ev)
```
caseydavenport: Hm, will this append do the right thing if and env var with the same `name` but different `value` is already present in the environment?

PR #324, opened 2019-09-10
Add semaphore.yml

neiljerram:
Use Semaphore v2 to run the K8ST test, as that test suite requires resources that are on the edge of what Semaphore v1 can provide.


neiljerram:
@caseydavenport If you think this looks right, I presume I need to apply it to release-v3.x branches as well, and then we can delete the k8st job from the Semaphore v1 settings.  How far back in v3.x should I go?

PR #329, opened 2019-09-14
checkFelixHealth remove named return

xieyanker:
## Description

bug fix

if the function of `resp.Body.Close()` get error not nil, it will return the error about closing failed. But in my opinion,the felix maybe normal, it should not return the error about closing failed.

## Release Note

```release-note
None required
```


caseydavenport:
Yeah seems fine to me. Thanks for the PR!

PR #333, opened 2019-10-01
Update node k8st's for new BGPConfig serviceExternalIPs and serviceClusterIPs fields

lmm:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/tests/k8st/tests/test_bgp_advert.py
+++ b/tests/k8st/tests/test_bgp_advert.py
@@ -393,7 +393,7 @@ spec:
 EOF
 """)
 
-        # Assert that a route to the service IP range is present.
+            # Assert that a route to the service IP range is present.
             retry_until_success(lambda: self.assertIn("10.96.0.0/12", self.get_routes()))
 
             # Create both a Local and a Cluster type NodePort service with a single replica.
--- a/tests/k8st/tests/test_bgp_advert.py
+++ b/tests/k8st/tests/test_bgp_advert.py
@@ -286,6 +286,8 @@ EOF
         self.tearDown()
         self.setUpRR()
 
+        localSvcExternalIP = "8.8.8.8"
+
         # Create ExternalTrafficPolicy Local service with one endpoint on node-1
         kubectl("""apply -f - << EOF
 apiVersion: apps/v1
@@ -333,8 +335,10 @@ spec:
     run: nginx-rr
   type: NodePort
   externalTrafficPolicy: Local
+  externalIPs:
+  - %s
 EOF
-""")
+""" % localSvcExternalIP)
 
         calicoctl("get nodes -o yaml")
         calicoctl("get bgppeers -o yaml")
@@ -351,7 +355,8 @@ EOF
 """ % json.dumps(node_dict))
 
         # Disable node-to-node mesh and configure bgp peering
-        # between node-1 and RR and also between external node and RR
+        # between node-1 and RR and also between external node and RR.
+        # Enable clusterIP and externalIPs advertisement.
         calicoctl("""apply -f - << EOF
 apiVersion: projectcalico.org/v3
 kind: BGPConfiguration
@@ -359,8 +364,13 @@ metadata: {name: default}
 spec:
   nodeToNodeMeshEnabled: false
   asNumber: 64512
+  serviceClusterIPs:
+  - cidr: 10.96.0.0/12
+  serviceExternalIPs:
+  - cidr: %s/32
 EOF
-""")
+""" % localSvcExternalIP)
+
         calicoctl("""apply -f - << EOF
 apiVersion: projectcalico.org/v3
 kind: BGPPeer
@@ -371,10 +381,15 @@ spec:
   asNumber: 64512
 EOF
 """)
+        # Verify clusterIP is advertised.
         svc_json = kubectl("get svc nginx-rr -n bgp-test -o json")
         svc_dict = json.loads(svc_json)
-        svcRoute = svc_dict['spec']['clusterIP']
-        retry_until_success(lambda: self.assertIn(svcRoute, self.get_routes()))
+        clusterIP = svc_dict['spec']['clusterIP']
+        retry_until_success(lambda: self.assertIn(clusterIP, self.get_routes()))
+
+        # Verify externalIP is advertised.
+        local_svc_externalips_route = "%s via 10.192.0.3" % localSvcExternalIP
+        retry_until_success(lambda: self.assertIn(local_svc_externalips_route, self.get_routes()))
 
     def test_mainline(self):
         """
--- a/tests/k8st/tests/test_bgp_advert.py
+++ b/tests/k8st/tests/test_bgp_advert.py
@@ -564,6 +564,18 @@ EOF
         Creates a lot of services quickly
         """
         with DiagsCollector():
+
+            calicoctl("""apply -f - << EOF
+apiVersion: projectcalico.org/v3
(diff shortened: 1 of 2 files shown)


tests/k8st/tests/test_bgp_advert.py
```diff
@@ -263,10 +273,21 @@ def assert_ecmp_routes(self, dst, via=["10.192.0.3", "10.192.0.4"]):
...
+    def add_svc_external_ips(self, svc, ns, ips):
+        ipsStr = ','.join('"{0}"'.format(ip) for ip in ips)
+        patchStr = "{\"spec\": {\"externalIPs\": [%s]}}" % (ipsStr)
+        return kubectl("patch svc %s -n %s --patch '%s'" % (svc, ns, patchStr)).strip()
+
     def test_rr(self):
         self.tearDown()
         self.setUpRR()
 
+        localSvcExternalIP = "8.8.8.8"
```
caseydavenport: I'd recommend using something other than `8.8.8.8`, since that's a commonly used DNS server.

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -352,17 +381,25 @@ def test_rr(self):
...
+        local_svc_externalips_route = "%s via 10.192.0.3" % localSvcExternalIP
+        retry_until_success(lambda: self.assertIn(local_svc_externalips_route, self.get_routes()))
 
     def test_mainline(self):
         """
         Runs the mainline tests for service ip advertisement
         - Create both a Local and a Cluster type NodePort service with a single replica.
           - assert only local and cluster CIDR routes are advertised.
           - assert /32 routes are used, source IP is preserved.
+          - add an external IP to both services.
```
caseydavenport: I'd recommend splitting this into two tests - one for clusterIP advertisement, and one for external IP advertisement.

Easy to understand breakages when the tests are small and focused.

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -430,20 +467,64 @@ def test_mainline(self):
...
+            calicoctl("""apply -f - << EOF
+apiVersion: projectcalico.org/v3
+kind: BGPConfiguration
+metadata:
+  name: default
+spec:
+  serviceClusterIPs:
+  - cidr: 10.96.0.0/12
+  serviceExternalIPs:
+  - cidr: %s/32
```
caseydavenport: Looks like these tests only test `/32` addresses here. I suspect most people will use a range.

e.g., allocate all of `172.24.16.0/24` for external IPs and then assign them out from that.

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -160,12 +160,6 @@ def setUp(self):
 
         start_external_node_with_bgp("kube-node-extra", bird_conf)
 
-        # set CALICO_ADVERTISE_CLUSTER_IPS=10.96.0.0/12
```
caseydavenport: Do we not need to replace this with a call to set up the BGPConfiguration?

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -354,21 +366,34 @@ def test_rr(self):
...
+kind: BGPConfiguration
+metadata:
+  name: default
+spec:
+  serviceClusterIPs:
+  - cidr: 10.96.0.0/12
+EOF
+""")
+
+        # Assert that a route to the service IP range is present.
```
caseydavenport: Indentation off here

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -354,21 +366,34 @@ def test_rr(self):
...
           - assert only local and cluster CIDR routes are advertised.
           - assert /32 routes are used, source IP is preserved.
-        - Create a local LoadBalancer service with clusterIP = None, assert no change.
         - Scale the Local NP service so it is running on multiple nodes, assert ECMP routing, source IP is preserved.
         - Delete both services, assert only cluster CIDR route is advertised.
         """
         with DiagsCollector():
-            # Assert that a route to the service IP range is present.
+
+            calicoctl("""apply -f - << EOF
```
caseydavenport: Do we need to delete this BGPConfiguration after the test? (Or maybe we already have something that does that?)

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -441,14 +466,122 @@ def test_mainline(self):
...
+            # Get clusterIPs.
+            local_svc_ip = self.get_svc_cluster_ip(local_svc, self.ns)
+            cluster_svc_ip = self.get_svc_cluster_ip(cluster_svc, self.ns)
+
+            # Wait for the deployments to roll out.
+            self.wait_for_deployment(local_svc, self.ns)
+            self.wait_for_deployment(cluster_svc, self.ns)
+
+            # Assert that clusterIPs are not advertised.
+            retry_until_success(lambda: self.assertNotIn(local_svc_ip, self.get_routes()))
```
caseydavenport: Shouldn't the `Local` type one be advertised?

caseydavenport: Oh wait nevermind, we don't set `clusterIPs` in the BGP config for this one.

tests/k8st/tests/test_bgp_advert.py
lmm: I moved this to the start of each of the non-rr tests. I considered making a function out of this to template the clusterIP and externalIP fields but not sure if it's worth it or not.

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -441,14 +466,122 @@ def test_mainline(self):
...
+    - ipBlock: { cidr: 10.192.0.5/32 }
+    ports:
+    - protocol: TCP
+      port: 80
+EOF
+""")
+
+            # Get host IPs for the nginx pods.
+            local_svc_host_ip = self.get_svc_host_ip(local_svc, self.ns)
+            clusterSvcHostIp = self.get_svc_host_ip(cluster_svc, self.ns)
```
caseydavenport: `cluster_svc_host_ip`?

tests/k8st/tests/test_bgp_advert.py
lmm: The bgpconfig is reset in tearDown

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -441,14 +466,122 @@ def test_mainline(self):
...
+            retry_until_success(lambda: self.assertNotIn(cluster_svc_externalips_route, self.get_routes()))
+
+            # Scale the local_svc to 4 replicas.
+            self.scale_deployment(local_svc, self.ns, 4)
+            self.wait_for_deployment(local_svc, self.ns)
+
+            # Verify that we have ECMP routes for the external IP of the local service.
+            retry_until_success(lambda: self.assert_ecmp_routes(local_svc_external_ip))
+
+            # Delete both services, assert only cluster CIDR route is advertised.
```
caseydavenport: I think this comment is wrong.

tests/k8st/tests/test_bgp_advert.py
caseydavenport: I see you've done it in the individual tests. Makes sense, since we have one test that doesn't use clusterIPs.

PR #336, opened 2019-10-05
Fix local build

Jc2k:
## Description

Tried to build a test image when looking in to #429 and found setting `LOCAL_BUILD` didn't result in a working build. This turned out to be because `EXTRA_DOCKER_ARGS` is modified after `DOCKER_RUN` is defined. Don't know if this works for some versions of make but it doesn't work for GNU Make 3.81 on macOS.

Attached patch defines `DOCKER_RUN` after the local build code has set `EXTRA_DOCKER_ARGS`, which fixes the build.
 
## Todos

Just a fix to dev tooling so I don't think these apply?

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -170,18 +170,6 @@ endif
 
 EXTRA_DOCKER_ARGS += -v $(GOMOD_CACHE):/go/pkg/mod:rw
 
-DOCKER_RUN := mkdir -p .go-pkg-cache $(GOMOD_CACHE) && \
-        docker run --rm \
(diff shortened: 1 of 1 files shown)


caseydavenport:
@rafaelvanoni was looking into this as well.

caseydavenport:
Thanks for the PR - looks like this has already been addressed in master, though.

PR #339, opened 2019-10-08
Make blocksize configurable for default ippool

halfcrazy:
fix invalid pseuod version in go.mod
fix https://github.com/projectcalico/calico/issues/2627

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


tmjd:
@halfcrazy Could you add reference documentation for these new environment variables?
It would need to go in https://github.com/projectcalico/calico/blob/master/master/reference/node/configuration.md
And would show up here https://docs.projectcalico.org/v3.9/reference/node/configuration

olljanat:
@halfcrazy I wonder that why did you add hard limit which prevents bigger block sizes than with IPv4 /20 or IPv6 /116 which mean only 4096 IPs per nodes to be used?

I want build up future proof IPv6 only Kubernetes cluster and was planning to use at least /112 (65536 IPs) block size on it before I noticed this limit.

halfcrazy:
> @halfcrazy I wonder that why did you add hard limit which prevents bigger block sizes than with IPv4 /20 or IPv6 /116 which mean only 4096 IPs per nodes to be used?
> 
> I want build up future proof IPv6 only Kubernetes cluster and was planning to use at least /112 (65536 IPs) block size on it before I noticed this limit.
> 
> EDIT: Opened #1337 to change default.

Blocksize is not designed to be too larger since a node could have multiple blocks, it will auto-scale up.

PR #344, opened 2019-10-10
Enable node to update CNI kubeconfig as needed

caseydavenport:
When using projected serviceaccount tokens, the kubelet will rotate the token file periodically. This invalidates the one in use by the Calico CNI plugin.

This PR enabled node to detect when the token has changed, and update the CNI plugin's kubeconfig file with the new information in response so long as the following is true:

- CALICO_MANAGE_CNI is set
- The CNI network directory is mounted into calico/node. 

- [ ] Tests
- [ ] Documentation
- [ ] Release note

```release-note
calico/node now updates CNI kubeconfig when credentials change
```

### DIFF
--- a/pkg/cni/token_watch.go
+++ b/pkg/cni/token_watch.go
@@ -1,0 +1,103 @@
+package cni
+
+import (
+	"encoding/base64"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"time"
+
+	"github.com/sirupsen/logrus"
+	"gopkg.in/fsnotify/fsnotify.v1"
+	"k8s.io/client-go/rest"
+)
+
+var serviceaccountDirectory string = "/var/run/secrets/kubernetes.io/serviceaccount/"
+var kubeconfigPath string = "/host/etc/cni/net.d/calico-kubeconfig"
+
+func Run() {
+	// Log to stdout.  this prevents our logs from being interpreted as errors by, for example,
+	// fluentd's default configuration.
+	logrus.SetOutput(os.Stdout)
+
+	// Create a watcher for file changes.
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		panic(err)
+	}
+	defer watcher.Close()
+
+	// Watch for changes to the serviceaccount directory. Rety if necessary.
+	for {
+		if err := watcher.Add(serviceaccountDirectory); err != nil {
+			// Error watching the file - retry
+			logrus.WithError(err).Error("Failed to watch Kubernetes serviceaccount files.")
+			time.Sleep(5 * time.Second)
+			continue
+		}
+
+		// Successfully watched the file. Break from the retry loop.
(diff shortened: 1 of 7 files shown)


pkg/cni/token_watch.go
```diff
@@ -0,0 +1,98 @@
...
+				err = rest.LoadTLSFiles(cfg)
+				if err != nil {
+					logrus.WithError(err).Error("Error loading TLS files")
+					continue
+				}
+				writeKubeconfig(cfg)
+
+			case err := <-watcher.Errors:
+				// We've received an error - log it out but don't exit.
+				logrus.WithError(err).Error("Error watching serviceaccount files")
```
tmjd: Should we have any type of delay here incase this just keeps firing?

pkg/cni/token_watch.go
```diff
@@ -0,0 +1,98 @@
...
+			}
+		}
+	}()
+
+	// Watch for changes to the serviceaccount token file.
+	for {
```
tmjd: Couldn't we just do this for before the go routine and do away with needing to run a go routine or reading from the stop below?

caseydavenport: Yeah I'll give that a try. This was done basically just by following the fsnotify examples :)

caseydavenport:
> This feels odd to me when we use to have a CNI pod that ran to monitor certificates. Do we just need that re-instated to monitor this also?

@tmjd We could, fundamentally. I think this is simpler and nicer (the old approach was some janky bash script). The old approach was also just for etcd secrets, not the k8s API access credentials.

pkg/cni/token_watch.go
```diff
@@ -0,0 +1,103 @@
...
+	"strings"
+	"time"
+
+	"github.com/sirupsen/logrus"
+	"gopkg.in/fsnotify/fsnotify.v1"
+	"k8s.io/client-go/rest"
+)
+
+var serviceaccountDirectory string = "/var/run/secrets/kubernetes.io/serviceaccount/"
+var kubeconfigPath string = "/host/etc/cni/net.d/calico-kubeconfig"
```
tmjd: I'm concerned this is hard coded. But this is going to be dependent on how we mount the directory so maybe that is fine.

pkg/cni/token_watch.go
```diff
@@ -0,0 +1,103 @@
...
+- name: calico-context
+  context:
+    cluster: local
+    user: calico
+current-context: calico-context`
+
+	// Replace the placeholders.
+	data = strings.Replace(data, "HOST", cfg.Host, 1)
+	data = strings.Replace(data, "TOKEN", cfg.BearerToken, 1)
+	data = strings.Replace(data, "CADATA", string(base64.StdEncoding.EncodeToString(cfg.CAData)), 1)
```
tmjd: Is there a chance that one of the previous replacements would have CADATA in the replacement? What about just doing an Sprintf instead of the 3 separate replacements?

PR #351, opened 2019-10-17
Run an apt-upgrade to make sure we're using the latest packages

caseydavenport:
- [ ] Tests
- [ ] Documentation
- [ ] Release note

```release-note
Make sure we include latest packages
```

### DIFF
--- a/Dockerfile.amd64
+++ b/Dockerfile.amd64
@@ -22,7 +22,7 @@ LABEL maintainer "Casey Davenport <casey@tigera.io>"
 ARG ARCH=amd64
 
 # Install remaining runtime deps required for felix from the global repository
-RUN apt-get update && apt-get install -y \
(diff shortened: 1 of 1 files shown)


Dockerfile.amd64
```diff
@@ -22,7 +22,7 @@ LABEL maintainer "Casey Davenport <casey@tigera.io>"
 ARG ARCH=amd64
 
 # Install remaining runtime deps required for felix from the global repository
-RUN apt-get update && apt-get install -y \
+RUN apt-get update && apt-get upgrade && apt-get install -y \
```
bcreane: if at first you don't succeed?

caseydavenport: Naw, `update` and `upgrade` do different things :)

PR #360, opened 2019-10-26
code optimization for read str's prefix

xieyanker:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/cni/token_watch.go
+++ b/pkg/cni/token_watch.go
@@ -27,7 +27,7 @@ func Run() {
 	}
 	defer watcher.Close()
 
-	// Watch for changes to the serviceaccount directory. Rety if necessary.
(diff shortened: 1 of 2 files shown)


xieyanker:
assign @caseydavenport

xieyanker:
@caseydavenport  Yes, it's not a bug, but I find the function `strings.HasPrefix` will be called serval times. It seems that not very efficient.
Of course, my code may not be very useful. If it's not necessary, you can close this PR.

caseydavenport:
I think in this code path that's OK - it's called infrequently and is not critical path. The calls to `HasPrefix` will be inconsequential. 

Thanks very much for the PR. However I think we should close this since it's mostly cosmetic.

PR #362, opened 2019-10-28
Use ubi-minimal for amd64 image

lmm:
## Description

This PR updates the amd64 Dockerfile to use `registry.access.redhat.com/ubi8/ubi-minimal:latest` as  the base image.

Some notes:
* Part of the complexity of the updated dockerfile has to do with the rebuilding of the iptables rpm. This is needed because the iptables packages in current ubi, rhel8, and CentOS repos have stripped out the legacy xtables-based iptables binaries.
* runit is not available in ubi, rhel8 or CentOS repos hence building from source
* `conntrack-tools` is not available in the ubi repo so for consistency and maybe better compatibility, we're using CentOS packages exclusively.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Calico component images are now using ubi for the base image
```

### DIFF
--- a/Dockerfile.amd64
+++ b/Dockerfile.amd64
@@ -16,79 +16,93 @@ ARG BIRD_IMAGE=calico/bird:latest
 FROM calico/bpftool:v5.0-amd64 as bpftool
 FROM ${BIRD_IMAGE} as bird
 
+# Use this build stage to build iptables rpm and runit binaries.
+# We need to rebuild the iptables rpm because the prepackaged rpm does not have legacy iptables binaries.
+# We need to build runit because there aren't any rpms for it in our repos.
 FROM centos:8 as centos
+
 ARG ARCH=x86_64
-ARG CDN_BASE_URL=https://cdn-ubi.redhat.com/content/public/ubi/dist/ubi8/8/${ARCH}/
+ARG CENTOS_MIRROR_BASE_URL=http://vault.centos.org/8.0.1905
 ARG LIBNFTNL_VER=1.1.1-4
-ARG LIBNFTNL_SOURCERPM_URL=${CDN_BASE_URL}/baseos/source/SRPMS/Packages/l/libnftnl-${LIBNFTNL_VER}.el8.src.rpm
-ARG LIBNFTNL_RPM_URL=${CDN_BASE_URL}/baseos/os/Packages/l/libnftnl-${LIBNFTNL_VER}.el8.${ARCH}.rpm
+ARG LIBNFTNL_SOURCERPM_URL=${CENTOS_MIRROR_BASE_URL}/BaseOS/Source/SPackages/libnftnl-${LIBNFTNL_VER}.el8.src.rpm
 ARG IPTABLES_VER=1.8.2-9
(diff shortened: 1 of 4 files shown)


Dockerfile.amd64
```diff
@@ -16,52 +16,147 @@ ARG BIRD_IMAGE=calico/bird:latest
...
+    dnf config-manager --set-enabled PowerTools && \
+    # Install required packages for building rpms. yum-utils is not required but it gives us yum-builddep to easily install build deps.
+    yum install -y rpm-build yum-utils make && \
+    # Need these to build runit.
+    yum install -y wget glibc-static gcc && \
+    # Ensure security updates are installed.
+    yum -y update-minimal --security --sec-severity=Important --sec-severity=Critical
+
+# In order to rebuild the iptables RPM, we first need to rebuild the libnftnl RPM because building
+# iptables requires libnftnl-devel but libnftnl-devel is not available on ubi or CentOS repos.
```
neiljerram: Is libnftnl-devel really not available even with PowerTools enabled?

Dockerfile.amd64
```diff
@@ -16,52 +16,147 @@ ARG BIRD_IMAGE=calico/bird:latest
...
+# Copy in runit binaries
+COPY --from=centos  /tmp/admin/runit-${RUNIT_VER}/command/* /usr/local/bin/
+
+# Copy in our rpms
+COPY --from=centos  /root/rpmbuild/RPMS/${ARCH}/* /tmp/rpms/
+
+COPY centos.repo /etc/yum.repos.d/
+
+# Install everything but conntrack, making sure that we're using only CentOS repos.
+# We're using CentOS for all our packages for consistency and probably better compatibility.
```
neiljerram: What do you mean by "probably better compatibility" here?

Makefile
```diff
@@ -340,7 +340,7 @@ endif
 	docker run --rm -v $(CURDIR)/dist/bin:/go/bin:rw $(CALICO_BUILD) /bin/sh -c "\
 	  echo; echo calico-node-$(ARCH) -v;	 /go/bin/calico-node-$(ARCH) -v; \
 	"
-	docker build --pull -t $(BUILD_IMAGE):latest-$(ARCH) . --build-arg BIRD_IMAGE=$(BIRD_IMAGE) --build-arg QEMU_IMAGE=$(CALICO_BUILD) --build-arg ver=$(CALICO_GIT_VER) -f ./Dockerfile.$(ARCH)
+	docker build --pull -t $(BUILD_IMAGE):latest-$(ARCH) . --build-arg BIRD_IMAGE=$(BIRD_IMAGE) --build-arg QEMU_IMAGE=$(CALICO_BUILD) --build-arg GIT_VERSION=$(CALICO_GIT_VER) -f ./Dockerfile.$(ARCH)
```
neiljerram: Does the change from `ver` to `GIT_VERSION` mean you have to update the Dockerfiles for other arches?

Dockerfile.amd64
lmm: I wanted to convey that there is less chance of package dependency issues if all of our packages were from a single source (CentOS) rather than a mix of CentOS and ubi. I'll reword

PR #369, opened 2019-11-20
Adding service checks for confd, bird and bird6 on the liveness probe

asincu:
## Description
These checks will be enabled when using option -bird-live or -bird6-live for the liveness probe. The default will be false.

https://github.com/projectcalico/calico/issues/2889

## Todos
- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Adding -bird-live/-bird6-live option to check for bird, bird6 and conf as part of the liveness probe. 
```

### DIFF
--- a/pkg/health/health.go
+++ b/pkg/health/health.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"os/exec"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


pkg/health/health.go
```diff
@@ -55,6 +56,14 @@ func Run(bird, bird6, felixReady, felixLive bool, thresholdTime time.Duration) {
 		os.Exit(0)
 	}
 
+	if birdLive {
+		if err := checkBIRDLive(); err != nil {
+			fmt.Printf("calico/node is not ready: Bird/ConfD is not live: %+v", err)
+			os.Exit(1)
+		}
+		os.Exit(0)
```
matthewdupre: The Exit(0) above this (l56) will stop people from being able to check both Felix and BIRD readiness.  The flow through this function will need tweaking to allow (at least) any combination of liveness checks.

pkg/health/health.go
```diff
@@ -82,6 +91,40 @@ func Run(bird, bird6, felixReady, felixLive bool, thresholdTime time.Duration) {
 	}
 }
 
+func checkBIRDLive() error {
+	error := checkService("confd")
+	if error != nil {
+		return error
+	}
+
+	error = checkService("bird")
```
matthewdupre: I don't think we always run both bird and bird6.  Certainly the readiness implementation (with separate flags) suggests that.

pkg/health/health.go
```diff
@@ -55,6 +56,14 @@ func Run(bird, bird6, felixReady, felixLive bool, thresholdTime time.Duration) {
 		os.Exit(0)
 	}
 
+	if birdLive {
+		if err := checkBIRDLive(); err != nil {
+			fmt.Printf("calico/node is not ready: Bird/ConfD is not live: %+v", err)
```
caseydavenport: nit: `bird/confd` (consistent capitalization with other places)

pkg/health/health.go
asincu: Sounds good! Thanks for spotting this!

pkg/health/health.go
asincu: I will add extra flags for bird6.

PR #370, opened 2019-11-20
Move common Makefile targets to Makefile.common

rafaelvanoni:


Makefile
```diff
@@ -1,118 +1,64 @@
...
+WGET?=/usr/bin/wget
 
-# unless otherwise set, I am building for my own architecture, i.e. not cross-compiling
-ARCH ?= $(BUILDARCH)
+Makefile.common: Makefile.common.$(MAKE_BRANCH)
+	cp "$<" "$@"
+Makefile.common.$(MAKE_BRANCH): $(WGET)
+	# Clean up any files downloaded from other branches so they don't accumulate.
+	rm -f Makefile.common.*
+	$(WGET) -nv $(MAKE_REPO)/Makefile.common -O "$@"
```
ozdanborne: Why this instead of just relying on whatever is in `$PATH`?
```suggestion
	wget -nv $(MAKE_REPO)/Makefile.common -O "$@"
```

Makefile
```diff
@@ -1,118 +1,64 @@
...
-
-# canonicalized names for host architecture
-ifeq ($(BUILDARCH),aarch64)
-	BUILDARCH=arm64
-endif
-ifeq ($(BUILDARCH),x86_64)
-	BUILDARCH=amd64
-endif
+# Download and include Makefile.common
+#   Additions to EXTRA_DOCKER_ARGS need to happen before the include since
```
ozdanborne: Could we make this requirement go away by using a recrusively expanded variable in Makefile.common for docker_run (`DOCKER_RUN =` ) instead of a simply expanded variable (`DOCKER_RUN :=` )

Makefile
```diff
@@ -1,118 +1,64 @@
...
 
-# canonicalized names for target architecture
-ifeq ($(ARCH),aarch64)
-	override ARCH=arm64
-endif
-ifeq ($(ARCH),x86_64)
-	override ARCH=amd64
-endif
+# Build mounts for running in "local build" mode. This allows an easy build using local development code,
+# assuming that there is a local checkout of libcalico in the same directory as this repo.
```
ozdanborne: Nitpick, but this comment sort-of implies that they should be in this directory

```suggestion
# assuming that there is a local checkout of libcalico alongside this repo in a sibling directory
```

Makefile
```diff
@@ -1,118 +1,64 @@
...
+ifdef LOCAL_BUILD
+PHONY: set-up-local-build
+LOCAL_BUILD_DEP:=set-up-local-build
```
ozdanborne: What's this thing do?

Makefile
rafaelvanoni: Gaurav asked to add it as a dependency.

Makefile
rafaelvanoni: I tried, it didn't work.

Makefile
rafaelvanoni: :man_shrugging:

Makefile
rafaelvanoni: Shaun re-wrote the local build rule in Felix (but didn't push it to the other repos) so it depends on a variable instead of a phony target.

ozdanborne: I can't figure why that means we need this line here. This isn't the Felix repo, and I don't see any references to `LOCAL_BUILD_DEP` or `set-up-local-build` in the Makefile.common repo.

Makefile
ozdanborne: It'd be nice if we could add it as a dependency without requiring it exist at `/usr/bin/wget`. But maybe not worth the effort because for most (all?) systems it is there. Is it there for Mac OS too?

Makefile
rafaelvanoni: Good catch, the `build` target was missing that dependency.

rafaelvanoni: Ah.. no, it's there in the rule for `NODE_CONTAINER_BINARY`

Makefile
rafaelvanoni: Will check on where Mac OS puts it...

rafaelvanoni: Seems that Mac OS doesn't ship wget, so it's up to the user to install it. Should I move all of the repos to using curl(1) or just rely on the user to install/link to it in the right place? I've no preference on this tbh.

rafaelvanoni: Ok, I'm moving all repos to curl and removing the dependency with a separate set of PRs.

PR #387, opened 2019-12-17
fail fast & retry setting NetworkUnavailable to false

spikecurtis:
Signed-off-by: Spike Curtis <spike@tigera.io>

## Description

If kube-proxy isn't working when we start, we'll currently wait ~30 seconds to timeout. This changes to a 2 second timeout, retried until a total of 30 seconds.  This should address the case where kube-proxy is ready a few seconds after Calico starts.

## Todos
- [ ] Tests
- [x] Documentation not needed
- [ ] Release note

## Release Note

```release-note
Fail fast & retry setting NetworkUnavailable to false (issue #307 )
```

### DIFF
--- a/pkg/startup/startup.go
+++ b/pkg/startup/startup.go
@@ -150,9 +150,9 @@ func Run() {
 		// If running under kubernetes with secrets to call k8s API
 		if config, err := rest.InClusterConfig(); err == nil {
 			log.Info("Setting NetworkUnavailable to False")
-			err = setNodeNetworkUnavailableFalse(config, nodeName)
+			err = setNodeNetworkUnavailableFalse(*config, nodeName)
 			if err != nil {
-				log.WithError(err).Errorf("Unable to set NetworkUnavailable to False")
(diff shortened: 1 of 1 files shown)


pkg/startup/startup.go
```diff
@@ -1113,8 +1119,19 @@ func setNodeNetworkUnavailableFalse(config *rest.Config, nodeName string) error
...
+	for {
+		select {
+		case <-to:
+			err = fmt.Errorf("timed out patching node, last error was: %s", err.Error())
+			return err
+		default:
+			_, err = clientset.CoreV1().Nodes().PatchStatus(nodeName, patch)
+			if err != nil {
+				log.WithError(err).Warnf("Failed to set NetworkUnavailable to False; will retry")
+			}
```
caseydavenport: I think we should break the loop here if we're successful, right?

spikecurtis: Yes! (and, duh!, my bad.)

PR #406, opened 2020-01-09
k8st: use same BIRD container as we use for dual ToR testing

neiljerram:
This is one step on a path to consolidate our ST suites around a common approach using KIND.


song-jiang:
This PR is now been covered by https://github.com/projectcalico/node/pull/407. We can close this one.

PR #407, opened 2020-01-14
Migrate k8st onto kind dual stack cluster.

song-jiang:
## Description

Migrate k8st onto kind cluster.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


Makefile
```diff
@@ -284,51 +284,36 @@ st-checks:
...
-	    -v /home/$(USER)/.kubeadm-dind-cluster:/root/.kubeadm-dind-cluster \
+	    -v ${HOME}/.kube/kind-config-kind:/root/.kube/config \
+	    -v $(CURDIR)/kubectl:/root/bin/kubectl \
 	    --privileged \
 	    --net host \
-	$(TEST_CONTAINER_NAME) \
-	    sh -c 'cp /root/.kubeadm-dind-cluster/kubectl /bin/kubectl && ls -ltr /bin/kubectl && which kubectl && cd /code/tests/k8st && \
-		   nosetests $(K8ST_TO_RUN) -v --with-xunit --xunit-file="/code/report/k8s-tests.xml" --with-timer'
+	${TEST_CONTAINER_NAME} \
+	    sh -c 'echo "container started.." && cp /root/bin/kubectl /bin/kubectl && echo "kubectl copied." && \
```
neiljerram: nit: Could we change the `-v` above to `-v $(CURDIR)/kubectl:/bin/kubectl`, and so avoid the need for this `cp`?

Makefile
```diff
@@ -284,51 +284,36 @@ st-checks:
...
+	     cd /code/tests/k8st &&  nosetests $(K8ST_TO_RUN) -s --nocapture --nologcapture -v --with-xunit --xunit-file="/code/report/k8s-tests.xml" --with-timer'
```
neiljerram: I guess we should now remove `-s --nocapture --nologcapture`

tests/k8st/create_kind_cluster.sh
```diff
@@ -0,0 +1,148 @@
...
+EOF
+  sed -i -e 's/"type": "calico-ipam"/"type": "calico-ipam",/' "${yaml}"
+
+  # And add all the IPV6 env vars
+  sed -i '/# Enable IPIP/r /dev/stdin' "${yaml}" << EOF
+            - name: IP6
+              value: "autodetect"
+            - name: CALICO_IPV6POOL_CIDR
+              value: "fd00:10:244::/64"
+            - name: CALICO_IPV6POOL_IPIP
```
neiljerram: Actually don't need CALICO_IPV6POOL_IPIP. (I have corrected the doc.)

tests/k8st/create_kind_cluster.sh
```diff
@@ -0,0 +1,148 @@
...
+              value: "Never"
+            - name: CALICO_IPV6POOL_NAT_OUTGOING
```
neiljerram: Actually don't need CALICO_IPV6POOL_NAT_OUTGOING.  (I have corrected the doc.)

tests/k8st/infra/calico-kdd.yaml
```diff
@@ -396,7 +414,7 @@ rules:
       - get
       - list
       - watch
-  # These permissions are only required for upgrade from v2.6, and can
+  # These permissions are only requried for upgrade from v2.6, and can
```
neiljerram: nit: Change here introduces a typo, so revert?

tests/k8st/create_kind_cluster.sh
```diff
@@ -0,0 +1,148 @@
...
+    # This happens when no matching resources exist yet,
+    # i.e. immediately after application of the Calico YAML.
+    sleep 5
+done
+echo "Calico is running."
+echo
+
+# Create and monitor a test webserver service for dual stack. 
+echo "Create test-webserver deployment..."
+kubectl apply -f tests/k8st/infra/test-webserver.yaml
```
neiljerram: Do we need this for the k8st tests?  If not, I think better to delete this and the associated YAML files.

tests/k8st/create_kind_cluster.sh
```diff
@@ -0,0 +1,148 @@
...
+networking:
+  disableDefaultCNI: true
+  podSubnet: "192.168.0.0/16,fd00:10:244::/64"
+  ipFamily: DualStack
+nodes:
+# the control plane node
+- role: control-plane
+- role: worker
+- role: worker
+- role: worker
```
neiljerram: How about just 3 nodes (i.e. control_plane + 2 workers), so that the setup is more similar to what we had with DIND?  Then there will be fewer changes in the BIRD config below.

tests/k8st/create_kind_cluster.sh
```diff
@@ -0,0 +1,148 @@
...
+  fi
+done
+echo
+
+echo "Download kind executable with dual stack support"
+# We need to replace kind executable and node image 
+# with official release once dual stack is fully supported by upstream.
+curl -L https://github.com/song-jiang/kind/releases/download/dualstack-1.17.0/kind -o ${KIND}
+chmod +x ${KIND}
+exit 0
```
neiljerram: Need to remove this `exit 0`

tests/k8st/create_kind_cluster.sh
song-jiang: Added connection tests to make sure kind cluster function correctly in terms of dual stack support.

tests/k8st/create_kind_cluster.sh
song-jiang: I feel we could keep another worker node here. By doing this, we keep the same configuration as our CI setup. Future test cases may requires 3 worker nodes but more than that is unlikely. There is only 3 lines code added to BIRD config to support it.

neiljerram: OK, fine.

PR #414, opened 2020-01-20
Support for IPv6 service advertisement

neiljerram:


neiljerram:
Docs and release note are at https://github.com/projectcalico/calico/pull/3127

tests/k8st/tests/test_bgp_advert_v6.py
```diff
@@ -0,0 +1,561 @@
+# Copyright (c) 2018 Tigera, Inc. All rights reserved.
```
song-jiang: Should be `2020 Tigera`

tests/k8st/tests/test_bgp_advert_v6.py
```diff
@@ -0,0 +1,561 @@
...
+class TestBGPAdvertV6(TestBaseV6):
+
+    def setUp(self):
+        super(TestBGPAdvertV6, self).setUp()
+
+        # Create bgp test namespace
+        self.ns = "bgp-test"
+        self.create_namespace(self.ns)
+
+        self.nodes, self.ip4s, self.ips = node_info()
```
song-jiang: I think we probably should use `ipv6` in this case.

tests/k8st/utils/utils.py
```diff
@@ -198,18 +223,28 @@ def generate_unique_id(length, prefix=""):
     random_string = ''.join(random.choice(string.ascii_lowercase + string.digits) for _ in range(length))
     return "%s-%s" % (prefix, random_string)
 
+ipv6_map = {
+    "kind-control-plane": "2001:20::8",
+    "kind-worker": "2001:20::1",
+    "kind-worker2": "2001:20::2",
+    "kind-worker3": "2001:20::3",
+}
+
```
song-jiang: Could you add comments that static map has to be used for ipv6 host ips , because kubernetes currently does not add it to host resources yet.

neiljerram:
Many thanks @song-jiang .  I've applied markups, now just waiting for CI to pass.

PR #415, opened 2020-01-20
[wip] Use new BIRD container

neiljerram:


neiljerram:
Will be included in #414

PR #416, opened 2020-01-20
Restructure BGP tests to avoid double setup for RR test

neiljerram:
This is a similar restructuring as I did a while back in node-private
as part of https://github.com/tigera/node-private/pull/112.
Specifically it avoids the RR test doing:

  - setup (for a non-RR test)
  - immediately tear that down again
  - setup for an RR test

which IIUC can lead to a rare flake.

### DIFF
--- a/tests/k8st/tests/test_bgp_advert.py
+++ b/tests/k8st/tests/test_bgp_advert.py
@@ -92,10 +92,10 @@ protocol bgp Mesh_with_node_2 from bgp_template {
 }
 """
 
-class TestBGPAdvert(TestBase):
+class _TestBGPAdvert(TestBase):
 
     def setUp(self):
-        super(TestBGPAdvert, self).setUp()
+        super(_TestBGPAdvert, self).setUp()
 
         # Create bgp test namespace
         self.ns = "bgp-test"
@@ -104,7 +104,7 @@ class TestBGPAdvert(TestBase):
         self.nodes, self.ips, _ = node_info()
         self.external_node_ip = start_external_node_with_bgp(
             "kube-node-extra",
-            bird_peer_config=bird_conf % (self.ips[0], self.ips[1], self.ips[2], self.ips[3])
+            bird_peer_config=self.get_bird_conf(),
         )
 
         # Enable debug logging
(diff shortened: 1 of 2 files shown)


tests/k8st/tests/test_bgp_advert.py
```diff
@@ -559,3 +448,133 @@ def check_routes_gone():
...
+    #                                      RR
+    # kube-master     kube-node-1     kube-node-2    kube-node-extra
+    #  10.192.0.2      10.192.0.3      10.192.0.4      10.192.0.5
+    #        |                |         | |    |         |
+    #        |                +---------+ |    +---------+
+    #        +----------------------------+   Peering -> is configured
+    #           These peerings are            by get_extra_peer_spec().
+    #           configured by BGPPeer         Peering <- is configured
+    #           peer-with-rr                  in get_bird_conf().
+
```
song-jiang: Nice!

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -211,109 +176,31 @@ def add_svc_external_ips(self, svc, ns, ips):
...
-""")
+    def get_bird_conf(self):
+        return bird_conf % (self.ips[0], self.ips[1], self.ips[2], self.ips[3])
 
-        calicoctl("""apply -f - << EOF
-apiVersion: projectcalico.org/v3
-kind: BGPPeer
-metadata: {name: kube-node-1}
+    def get_extra_peer_spec(self):
+        return """
```
song-jiang: What this is for?

neiljerram: It specifies the BGPPeer for the peering(s) to the external (i.e. non-cluster) node, and allows that to be different for the two test classes, i.e. the non-RR and the RR class.
- In the non-RR class, **all** the cluster nodes peer with the external node.
- In the RR class, the cluster nodes peer with the in-cluster RR node, and only the RR node peers with the external node.

Does that answer your question?

song-jiang: I see. Sorry I mistakenly thought all the functions just return "" but actually they are not.

PR #417, opened 2020-01-21
Auto-detect Pod CIDR on kubeadm

rafaelvanoni:
```release-note
Calico now auto-detects the IP Pool CIDR when running on kubeadm
```


pkg/startup/startup.go
```diff
@@ -720,6 +723,53 @@ func configureIPPools(ctx context.Context, client client.Interface) {
 		return
 	}
 
+	// If CIDRs weren't specified through the environment variables, check if they're present in kubeadm's
+	// config map.
+	if len(ipv4Pool) == 0 || len(ipv6Pool) == 0 {
+		clientset, err := createKubernetesClient()
+		if err == nil {
+			kubeadmConfig, err := clientset.CoreV1().ConfigMaps(metav1.NamespaceSystem).Get(KubeadmConfigConfigMap, metav1.GetOptions{})
+			if err != nil {
```
caseydavenport: I think we should handle two separate error cases:

- kubeadm configmap doesn't exist: this is OK and expected for a lot of environments, and shouldn't result in a warning log
- some other error: this is potentially an issue (e.g., failed to query the API server) and should result in a hard error, and probably should stop processing and exit. 

In the current iteration both of these cases are handled by a warning log.

pkg/startup/startup.go
```diff
@@ -720,6 +723,53 @@ func configureIPPools(ctx context.Context, client client.Interface) {
...
+
+				for _, line := range kubeadmConfig.Data {
+					match := re.FindStringSubmatch(line)
+					if match != nil {
+						found = true
+						// This string splitting assumes that IPv4 and IPv6 CIDRs can be
+						// entered in the same line as separated by an empty space.
+						for _, cidr := range strings.Split(match[1], " ") {
+							addr, _, err := net.ParseCIDR(cidr)
+
```
caseydavenport: nit: extra newline here

pkg/startup/startup.go
```diff
@@ -720,6 +723,53 @@ func configureIPPools(ctx context.Context, client client.Interface) {
...
+			} else {
+				// Look through the config map for lines starting with 'podSubnet', then assign
+				// the right variable according to the IP family of the matching string.
+				re := regexp.MustCompile(`podSubnet: (.*)`)
+				found := false
+
+				for _, line := range kubeadmConfig.Data {
+					match := re.FindStringSubmatch(line)
+					if match != nil {
+						found = true
```
caseydavenport: We generally try to leave a newline before comment blocks like this - it makes the code a bit easier to read. 

e.g.


```go
found = true

// Comment starts here
for _, cidr := range ... {
}
```

pkg/startup/startup.go
```diff
@@ -720,6 +723,53 @@ func configureIPPools(ctx context.Context, client client.Interface) {
...
+	if len(ipv4Pool) == 0 || len(ipv6Pool) == 0 {
+		clientset, err := createKubernetesClient()
+		if err == nil {
+			kubeadmConfig, err := clientset.CoreV1().ConfigMaps(metav1.NamespaceSystem).Get(KubeadmConfigConfigMap, metav1.GetOptions{})
+			if err != nil {
+				log.WithError(err).Warn("failed to query kubeadm's config map")
+			} else {
+				// Look through the config map for lines starting with 'podSubnet', then assign
+				// the right variable according to the IP family of the matching string.
+				re := regexp.MustCompile(`podSubnet: (.*)`)
```
caseydavenport: For my own understanding, with dual-stack this configmap will have multiple `podSubnet: ` lines in it?

go.sum
```diff
@@ -367,6 +367,8 @@ github.com/prometheus/procfs v0.0.0-20171017214025-a6e9df898b13 h1:leRfx9kcgnSDk
 github.com/prometheus/procfs v0.0.0-20171017214025-a6e9df898b13/go.mod h1:c3At6R/oaqEKCNdg8wHV1ftS6bRYblBhIjjI8uT2IGk=
 github.com/prometheus/procfs v0.0.0-20181005140218-185b4288413d h1:GoAlyOgbOEIFdaDqxJVlbOQ1DtGmZWs/Qau0hIlk+WQ=
 github.com/prometheus/procfs v0.0.0-20181005140218-185b4288413d/go.mod h1:c3At6R/oaqEKCNdg8wHV1ftS6bRYblBhIjjI8uT2IGk=
+github.com/rafaelvanoni/node v3.8.4+incompatible h1:279SEerpQxoMJNXqJvuybwjd0oVkYzHqynFUs895VHs=
```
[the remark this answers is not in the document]
rafaelvanoni: Fixed

pkg/startup/startup.go
```diff
@@ -148,15 +150,16 @@ func Run() {
...
 			log.Info("Setting NetworkUnavailable to False")
-			err = setNodeNetworkUnavailableFalse(*config, nodeName)
+			err = setNodeNetworkUnavailableFalse(*clientset, nodeName)
 			if err != nil {
 				log.WithError(err).Error("Unable to set NetworkUnavailable to False")
 			}
 		}
 	}
 
+	// attempt to discover the cluster type it is running on
```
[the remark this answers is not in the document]
rafaelvanoni: I copied from

> // configureNodeRef will attempt to discover the cluster type it is running on, check to ensure we
> // have not already set it on this Node, and set it if need be.
> func configureNodeRef(node *api.Node) {

pkg/startup/startup.go
```diff
@@ -705,7 +708,7 @@ func GenerateIPv6ULAPrefix() (string, error) {
 
 // configureIPPools ensures that default IP pools are created (unless explicitly
 // requested otherwise).
-func configureIPPools(ctx context.Context, client client.Interface) {
+func configureIPPools(ctx context.Context, client client.Interface, nodeName string) {
```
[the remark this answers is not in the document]
rafaelvanoni: Ah, yes... fixed

pkg/startup/startup.go
```diff
@@ -720,6 +723,53 @@ func configureIPPools(ctx context.Context, client client.Interface) {
 		return
 	}
 
+	// If CIDRs weren't specified through the environment variables, check if they're present in kubeadm's
+	// config map.
+	if len(ipv4Pool) == 0 || len(ipv6Pool) == 0 {
+		clientset, err := createKubernetesClient()
+		if err == nil {
```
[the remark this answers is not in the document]
rafaelvanoni: Fixed

pkg/startup/startup.go
```diff
@@ -148,15 +150,16 @@ func Run() {
 		configureASNumber(node)
 
 		// If running under kubernetes with secrets to call k8s API
-		if config, err := rest.InClusterConfig(); err == nil {
```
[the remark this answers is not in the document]
rafaelvanoni: It looks odd to init the client inside the branch and then have the following code assume it's set correctly... but sure.

pkg/startup/startup.go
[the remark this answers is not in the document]
caseydavenport: Ah, I think we're not being very precise in our language and using "cluster type" to mean two things. It's not the "cluster type", rather it determines the type of cluster (e.g, whether this is a Kubernetes cluster or something else). 

It's independent from the CLUSTER_TYPE as we're using it in this PR.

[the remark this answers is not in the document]
caseydavenport: If you feel a comment is really necessary for this line, then I'd say something like "Populate a reference to the node based on orchestrator node identifiers"

pkg/startup/startup.go
rafaelvanoni: Fixed

pkg/startup/startup.go
rafaelvanoni: Fixed

pkg/startup/startup.go
rafaelvanoni: Fixed

pkg/startup/startup.go
rafaelvanoni: No, from what I could find in related github issues it will be a single line with CIDRs separated by a comma. For example, see https://github.com/kubernetes/kubernetes/pull/82473

go.sum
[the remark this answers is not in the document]
rafaelvanoni: Fixed now

pkg/startup/startup.go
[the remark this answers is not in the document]
rafaelvanoni: Fixed

pkg/startup/startup.go
```diff
@@ -27,6 +27,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/juju/errors"
```
caseydavenport: Actually, I think this import is strange too - what is this library?

rafaelvanoni: Should be `k8s.io/apimachinery/pkg/api/errors` for `errors.IsNotFound` instead

pkg/startup/startup.go
```diff
@@ -1142,3 +1208,19 @@ func terminate() {
...
+
+// queryKubeadmConfigMap returns kubeadm's "kubeadm-config" config map if it exists.
+func queryKubeadmConfigMap(clientset *kubernetes.Clientset) (*kapiv1.ConfigMap, error) {
+	// clientset will only be initialized when running on k8s
+	if clientset == nil {
+		return nil, fmt.Errorf("Unable to read kubeadm config map, not running on kubernetes.")
+	}
+
+	kubeadmConfig, err := clientset.CoreV1().ConfigMaps(metav1.NamespaceSystem).Get(KubeadmConfigConfigMap, metav1.GetOptions{})
+	if err != nil {
```
caseydavenport: I think we need to handle `IsNotFound` explicitly here as well. Otherwise we'll log out an error on non-kubeadm systems.

pkg/startup/startup.go
```diff
@@ -720,6 +739,55 @@ func configureIPPools(ctx context.Context, client client.Interface) {
 		return
 	}
 
+	// If CIDRs weren't specified through the environment variables, check if they're present in kubeadm's
+	// config map.
+	if clientset != nil && (len(ipv4Pool) == 0 || len(ipv6Pool) == 0) {
+		kubeadmConfig, err := clientset.CoreV1().ConfigMaps(metav1.NamespaceSystem).Get(KubeadmConfigConfigMap, metav1.GetOptions{})
```
caseydavenport: Should we use the `queryKubeadmConfigMap` here?

rafaelvanoni: With the code review changes from earlier today `queryKubeadmConfigMap` became unnecessary. It's cleaner to call `Get()` directly on both use cases.

pkg/startup/startup.go
```diff
@@ -956,10 +1025,22 @@ func checkConflictingNodes(ctx context.Context, client client.Interface, node *a
 
 // ensureDefaultConfig ensures all of the required default settings are
 // configured.
-func ensureDefaultConfig(ctx context.Context, cfg *apiconfig.CalicoAPIConfig, c client.Interface, node *api.Node) error {
+func ensureDefaultConfig(ctx context.Context, cfg *apiconfig.CalicoAPIConfig, c client.Interface, node *api.Node, clientset *kubernetes.Clientset) error {
 	// Ensure the ClusterInformation is populated.
 	// Get the ClusterType from ENV var. This is set from the manifest.
 	clusterType := os.Getenv("CLUSTER_TYPE")
+
+	if clientset != nil {
```
caseydavenport: Just continuing to mull on this, what do you think about actually querying the configmap only once higher up in the call stack, and then passing that information down to the appropriate functions instead of the configmap?

I think it has some tiny benefits, namely fewer API calls (each of which might fail), slightly less code duplication, and makes UTs easier (no need to pass a mock clientset or set up cluster access, can just pass a configmap struct or a boolean). 

WDYT?

pkg/startup/startup_test.go
```diff
@@ -99,6 +100,19 @@ const (
 	randomULAPool = "<random ULA pool>"
 )
 
+// createClientset creates a kubernetes clientset when running under that platform.
+func createClientset() *kubernetes.Clientset {
```
caseydavenport: I don't think this is right - it relies on a real cluster being available (and also doesn't check for errors - it's just going to return a nil clientset IIUC). 

I think we should be using the fake client for UTs: https://godoc.org/k8s.io/client-go/kubernetes/fake

caseydavenport: (note my other comment above - I think we can remove the need to pass in a clientset to ensureDefaultConfig/configureIPPools simply by querying the configmap outside of those functions and passing in the configmap contents / boolean indicating if we're using kubeadm. that would remove dependence on the clientset here.)

pkg/startup/startup.go
```diff
@@ -720,6 +739,56 @@ func configureIPPools(ctx context.Context, client client.Interface) {
...
+								if len(ipv6Pool) == 0 {
+									ipv6Pool = cidr
+								}
+							} else {
+								if len(ipv4Pool) == 0 {
+									ipv4Pool = cidr
+								}
+							}
+						} else {
+							log.WithError(err).Warnf("failed to parse CIDR %v", cidr)
```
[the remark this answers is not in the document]
rafaelvanoni: Ok, I'll just remove `found`. It's not adding much value now that this chunk of code has been tested.

pkg/startup/startup.go
rafaelvanoni: It's the trade off between fewer API calls in the `kubeadm` case but unnecessary calls otherwise. I was leaning towards the former but open to changing that.

pkg/startup/startup.go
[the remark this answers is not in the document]
caseydavenport: I think there is value in having a log which indicates whether or not we've found addresses in the configmap, just that we don't need special logic for it.

pkg/startup/startup.go
caseydavenport: I don't think that's true - we always need to make one API call to query the configmap in order to determine if we're running on kubeadm.

In other words, right now we always query it twice. We could be querying it only once.

pkg/startup/startup.go
```diff
@@ -168,11 +189,13 @@ func Run() {
 		terminate()
 	}
 
+	kubeadmConfig, _ := queryKubeadmConfigMap(clientset)
```
caseydavenport: We definitely need to handle the error here. What happens if we try to query the configmap but hit an error connecting to or authorizing with the api server? I think we need to fail in that scenario.

The only errors we don't want to fail is:
- If we're not running on k8s
- If we get a "Does not exist" error.

pkg/startup/startup.go
```diff
@@ -720,6 +742,15 @@ func configureIPPools(ctx context.Context, client client.Interface) {
 		return
 	}
 
+	// If CIDRs weren't specified through the environment variables, check if they're present in kubeadm's
+	// config map.
+	if (len(ipv4Pool) == 0 || len(ipv6Pool) == 0) && kubeadmConfig != nil {
+		ipv4Pool, ipv6Pool, err := extractKubeadmCIDRs(kubeadmConfig)
+		if err == nil {
```
caseydavenport: Here's another case where we're not handling an error. We need an `err != nil` case here.

pkg/startup/startup.go
```diff
@@ -1142,3 +1170,63 @@ func terminate() {
...
+			terminate()
+			return nil, err
+		}
+	}
+
+	return cm, err
+}
+
+// extractKubeadmCIDRs looks through the config map and parses lines starting with 'podSubnet'.
+func extractKubeadmCIDRs(kubeadmConfig *v1.ConfigMap) (string, string, error) {
```
caseydavenport: I think we need UTs for this function, passing in a few different variations of the configmap and asserting we get the correct results.

pkg/startup/startup.go
rafaelvanoni: Check the implementation of `queryKubeadmConfigMap()`, the error handling for the API call is in there.

caseydavenport: Also, if we move this within InClusterConfig block above, then we won't need the "If clientset == nil" check within the function. It will let us group all of the Kuberentes-specific logic in one place.

caseydavenport: > Check the implementation of queryKubeadmConfigMap(), the error handling for the API call is in there.

But, it looks like the function returns an error but it's never checked anywhere, is that right?

pkg/startup/startup.go
```diff
@@ -720,6 +750,24 @@ func configureIPPools(ctx context.Context, client client.Interface) {
...
+			if len(ipv4Pool) == 0 {
+				ipv4Pool = v4
+				log.Infof("found v4=%s in the kubeadm config map", ipv4Pool)
+			}
+			if len(ipv6Pool) == 0 {
+				ipv6Pool = v6
+				log.Infof("found v6=%s in the kubeadm config map", ipv6Pool)
+			}
+		} else {
+			log.WithError(err).Error("Failed to extract CIDRs from kubeadm config.")
```
caseydavenport: If this is actually an error, then I think we need to exit. If this is not exit-worthy, then I think we should downgrade this log message to a warning level log.

(We try to keep error level logs to things that are actually problems the user needs to action. I'm not sure if this is one of those?)

pkg/startup/startup.go
```diff
@@ -1142,3 +1187,48 @@ func terminate() {
...
 	exitFunction(1)
 }
+
+// extractKubeadmCIDRs looks through the config map and parses lines starting with 'podSubnet'.
+func extractKubeadmCIDRs(kubeadmConfig *v1.ConfigMap) (string, string, error) {
+	var v4, v6 string
+	var line []string
+	var err error
+
+	if kubeadmConfig == nil {
```
caseydavenport: FWIW, it looks like we already check this prior to calling this function, so we probably don't need this check here. 

It's not wrong, per-se, but makes it less clear who is responsible for checking whether the configmap exists or not.

pkg/startup/startup_test.go
```diff
@@ -987,3 +989,23 @@ var _ = Describe("UT for GenerateIPv6ULAPrefix", func() {
...
+		}
+		Expect(v4).To(Equal(expectedIPv4))
+		Expect(v6).To(Equal(expectedIPv6))
+	},
+	Entry("nil config map", nil, "", "", true),
+	Entry("empty config map", &v1.ConfigMap{}, "", "", false),
+	Entry("v4 only config map", &v1.ConfigMap{Data: map[string]string{"ClusterConfiguration": "podSubnet: 192.168.0.0/16"}}, "192.168.0.0/16", "", false),
+	Entry("dual v4 config map", &v1.ConfigMap{Data: map[string]string{"ClusterConfiguration": "podSubnet: 192.168.0.0/16,10.10.0.0/16"}}, "192.168.0.0/16", "", false),
+	Entry("v6 only config map", &v1.ConfigMap{Data: map[string]string{"ClusterConfiguration": "podSubnet: fdff:ffff:ffff:ffff:ffff::/80"}}, "", "fdff:ffff:ffff:ffff:ffff::/80", false),
+	Entry("dual v6 config map", &v1.ConfigMap{Data: map[string]string{"ClusterConfiguration": "podSubnet: fdff:ffff:ffff:ffff:ffff::/80,fdff:ffff:ffff:ffff:ffff::/80"}}, "", "fdff:ffff:ffff:ffff:ffff::/80", false),
```
caseydavenport: I think it's worth at least one test with the full configmap present and not just the podSubnet.

e.g.,

```
ClusterConfiguration": "apiServerCertSANs:\n- 35.223.159.180\n- 127.0.0.1\napiServerExtraArgs:\n  audit-log-path: /var/log/calico/audit/kube-audit.log\n  audit-policy-file: /etc/kubernetes/pki/audit-policy.yaml\n  authorization-mode: Node,RBAC\n  basic-auth-file: /etc/kubernetes/pki/basic_auth.csv\napiServerExtraVolumes:\n- hostPath: /var/log/calico/audit/\n  mountPath: /var/log/calico/audit/\n  name: calico-audit\n  pathType: DirectoryOrCreate\n  writable: true\napiVersion: kubeadm.k8s.io/v1alpha3\nauditPolicy:\n  logDir: /var/log/kubernetes/audit\n  logMaxAge: 2\n  path: \"\"\ncertificatesDir: /etc/kubernetes/pki\nclusterName: kubernetes\ncontrolPlaneEndpoint: \"\"\netcd:\n  local:\n    dataDir: /var/lib/etcd\n    image: \"\"\nimageRepository: k8s.gcr.io\nkind: ClusterConfiguration\nkubernetesVersion: v1.12.7\nnetworking:\n  dnsDomain: cluster.local\n  podSubnet: 192.168.0.0/16\n  serviceSubnet: 10.96.0.0/12\nunifiedControlPlaneImage: \"\"\n",
```
(This is what my cluster gives me for `kubectl get cm -n kube-system kubeadm-config`)

pkg/startup/startup.go
rafaelvanoni: Yes.. If this was in C I would have put an assertion there :-) This is me trying to write self-documenting code so if anyone uses this routine in the future they'll know this argument can't be null.

caseydavenport: lol gotcha.

PR #418, opened 2020-01-27
add protocols section in nsswitch.conf

leodotcloud:
## Description

Type of fix: bug fix

Problem: Running calico in a NIS enabled environment, with no protocols info, crashes calico-node pod.

Root cause: ipsets command tries to resolve the protocol by first looking up nis, followed by files. In an environment where NIS is enabled but the server doesn't respond with the protocol information, the files (/etc/protocols) information is not used. Hence, the crash.

Fix: Specify protocols section in nsswitch.conf

Issue(s): https://github.com/rancher/rancher/issues/24543

```release-note
None required
```

### DIFF
--- a/filesystem/etc/nsswitch.conf
+++ b/filesystem/etc/nsswitch.conf
@@ -1,1 +1,2 @@
 hosts: files mdns4_minimal [NOTFOUND=return] dns mdns4
+protocols: files


caseydavenport:
This seems like it's probably OK. I wonder, will this impact other users who might be relying on the current behavior?

leodotcloud:
@caseydavenport This would impact users who rely on their NIS servers to reply to the protocols resolution request. Should we release note this?

leodotcloud:
If you inspect `/etc/protocols` file, most of the popular protocols are already covered. This fix would impact users if they are relying on their NIS servers to resolve protocols that are not listed in this file and those protocols are being used in the iptables/ipset rules. I am not sure how many users are relying on this behavior.

caseydavenport:
Thanks @leodotcloud!

leodotcloud:
Thanks @caseydavenport for the quick turnaround! Curious to know how often do you create a release? (so that we can pick up the latest changes)

caseydavenport:
We generally release patch releases only as needed, depending on the severity of bugs found and patched.

This change will go into v3.13.0, which we're targeting in mid-to-late Feb.

leodotcloud:
Appreciate the update, we will plan things accordingly on our end based on this timeline.

PR #420, opened 2020-02-06
Add RBAC for configmaps, for kubeadm CIDR auto-detection

neiljerram:


caseydavenport:
Approved, though there's a comment that says this section is just for the CNI plugin. Probably fine since it's just test code, but could be confusing since the CNI plugin doesn't need configmaps.

PR #421, opened 2020-02-10
Autoselect default IP pool if overlapping with host IPs

mhmxs:
## Description
Default IP pool can overlap with user's host network. This new feature introduced a logic to select different one on case of overlapping. On any case of error it fallbacks to the original default pool, so the change doesn't break current behavior. On case of overlapping calico node starts to find free IP range between `172.16.0.0/16` and `172.31.0.0/16`. This PR is related to https://github.com/projectcalico/calico/issues/1401. I have tested it manually and covered the change fully with automated test cases.

## Todos
- [ ] Tests
- [X] Documentation
- [x] Release note

## Release Note

```release-note
On case of overlapping between host network and default IP pool Calico node start to find a free IP range between `172.16.0.0/16` and `172.31.0.0/16`
```

### DIFF
--- a/pkg/startup/autodetection/ipv4/poolselector.go
+++ b/pkg/startup/autodetection/ipv4/poolselector.go
@@ -1,0 +1,83 @@
+// Copyright (c) 2020 Tigera, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package ipv4
+
+import (
+	"fmt"
+	"net"
+
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+	"github.com/vishvananda/netlink"
+)
+
+const (
+	FALLBACK_IPPOOL_TEMPLATE = "172.%d.0.0/16"
+	FALLBACK_IPPOOL_MIN      = 16
+	FALLBACK_IPPOOL_MAX      = 31
+)
+
+var hostIPAddressRetriever func(netlink.Link, int) ([]netlink.Addr, error) = netlink.AddrList
+
+// GetDefaultIPv4Pool detects host interfaces and selects default IP pool without overlapping
+func GetDefaultIPv4Pool(preferedPool *net.IPNet) (*net.IPNet, error) {
+	log.Debug("Auto-detecting IPv4 interfaces to select default pool")
+
+	hostAddrs, err := retrieveIPAddresses()
+	if err != nil {
+		log.Errorf("Unable to retrieve host network interfaces: '%s'", err.Error())
+		return nil, err
+	}
+
+	return findAvailableCIDR(preferedPool, hostAddrs), nil
+}
+
+func retrieveIPAddresses() ([]net.IPNet, error) {
+	linkAddresses, err := hostIPAddressRetriever(nil, netlink.FAMILY_V4)
+	if err != nil {
+		return nil, errors.Wrap(err, "Unable to list addresses")
+	}
+
+	resp := make([]net.IPNet, len(linkAddresses))
+	for _, addr := range linkAddresses {
+		resp = append(resp, *addr.IPNet)
+	}
+
+	return resp, nil
+}
+
+func findAvailableCIDR(preferredPool *net.IPNet, hostAddrs []net.IPNet) *net.IPNet {
+	if !doesPoolOverlap(preferredPool, hostAddrs) {
+		return preferredPool
+	}
+
+	for i := FALLBACK_IPPOOL_MIN; i <= FALLBACK_IPPOOL_MAX; i++ {
+		_, poolToTry, _ := net.ParseCIDR(fmt.Sprintf(FALLBACK_IPPOOL_TEMPLATE, i))
+		if !doesPoolOverlap(poolToTry, hostAddrs) {
+			return poolToTry
+		}
+	}
+
+	// We couldn't find a pool. For compatibility reasons, return the preferred pool to match existing behavior.
+	return preferredPool
+}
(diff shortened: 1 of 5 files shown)


mhmxs:
I was thinking in replacing command execution `ip -4 addr` to some Go native library. The down side is it requires to introduce new dependecy. But the advantage is `node` can use low level APIs to fetch network addresses. One of the candidates is : https://godoc.org/github.com/jsimonetti/rtnetlink. It has nice ListLink and ListAddresses methods.

The implementation made things a bit more complex than a simple exec.

List all interfaces 
```
// Dial a connection to the rtnetlink socket
conn, err := rtnetlink.Dial(nil)
if err != nil {
    log.Fatal(err)
}
defer conn.Close()

// Request a list of interfaces
msg, err := conn.Link.List()
if err != nil {
    log.Fatal(err)
}

log.Printf("%#v", msg)
```

and list addresses
```
// Gather the interface Index
iface, _ := net.InterfaceByName("lo")
// Get an ip address to add to the interface
family := uint8(unix.AF_INET)

// Dial a connection to the rtnetlink socket
conn, err := rtnetlink.Dial(nil)
if err != nil {
    log.Fatal(err)
}
defer conn.Close()

// Request a list of addresses
msg, err := conn.Address.List()
if err != nil {
    log.Fatal(err)
}

// Filter out the wanted messages and put them in the 'addr' slice.
var addr []rtnetlink.AddressMessage
for _, v := range msg {
    add := true
    if iface != nil && v.Index != uint32(iface.Index) {
        add = false
    }
    if family != 0 && v.Family != family {
        add = false
    }
    if add {
        addr = append(addr, v)
    }
}

log.Printf("%#v", addr)
```

Sources are the exmaple.

What is your advice?

caseydavenport:
@mhmxs thanks for the PR! 

I haven't had a chance to review in depth yet, but to your question - I think we already import https://github.com/vishvananda/netlink - could you use that to do what you're suggesting? I believe it support similar interface / address listing capabilities.

mhmxs:
@caseydavenport All right. I updated the PR to use netlink to determine host ip addresses. Code become cleaner.

pkg/startup/autodetection/ipv4/poolselector.go
```diff
@@ -0,0 +1,114 @@
...
+
+var hostIPAddressRetriever func() ([]net.IPNet, error) = retrieveIPAddresses
+
+// GetDefaultIPv4Pool detects host interfaces and selects default IP pool without overlapping
+func GetDefaultIPv4Pool(preferedPool string) string {
+	log.Debug("Auto-detecting IPv4 interfaces to select default pool")
+
+	_, preferedNet, err := net.ParseCIDR(preferedPool)
+	if err != nil {
+		log.Errorf("Unable to parse prefered pool: '%s'", err.Error())
```
caseydavenport: I think we should be returning errors from this function and having the calling code handle them. Presumably if, somehow, we fail to parse this CIDR, we don't want the calling code to continue on using the bad CIDR, right?

pkg/startup/autodetection/ipv4/poolselector.go
```diff
@@ -0,0 +1,114 @@
...
+
+func retrieveIPAddresses() ([]net.IPNet, error) {
+	linkAddresses, err := netlink.AddrList(nil, netlink.FAMILY_V4)
+	if err != nil {
+		return nil, errors.Wrap(err, "Unable to list addresses")
+	}
+
+	resp := make([]net.IPNet, len(linkAddresses))
+	for _, addr := range linkAddresses {
+		_, netAddr, err := net.ParseCIDR(addr.IPNet.String())
```
caseydavenport: Not a big deal, but do we actually need to ParseCIDR here?

Couldn't we just do `resp = append(resp, *addr.IPNet)`?

pkg/startup/autodetection/ipv4/poolselector.go
```diff
@@ -0,0 +1,114 @@
...
+	return resp, nil
+}
+
+func (s *poolSelector) selectDefaultPool() (string, error) {
+	for _, addr := range s.hostAddresses {
+		if s.actualPool.Contains(addr.IP) {
+			if err := s.hopToNextPool(); err != nil {
+				return "", errors.Wrap(err, "Unable to find unused IP pool")
+			}
+			return s.selectDefaultPool()
```
caseydavenport: I think the recursion here makes this code a bit harder to understand. How do you feel about unpicking this so that it doesn't use recursion?

Here's some pseudocode:

```go
func findAvailableCIDR(hostAddrs []net.IPNet) net.IPNet {
  // Check if the preferred pool overlaps with any host addresses. If it doesn't then just use that.
  if overlaps := doesPoolOverlap(preferredPool, hostAddrs); !overlaps {
    return preferredPool
  }

  // The preferred pool overlapped. Look for another one.
  for i = 16 -> i = 32 {
    pooltoTry = fmt.Sprintf(TEMPLATE, i)
    if overlaps = doesPoolOverlap(pooltoTrye, hostAddrs); !overlaps {
      // Found one, use it.
      return poolToTry
    }
  }

  // We couldn't find a pool. For compatibility reasons, return the preferred pool to match existing behavior.
  return preferredPool
}
```

pkg/startup/autodetection/ipv4/poolselector.go
mhmxs: Nice catch, during refactor original code i missed that in new code types are match :) Thanks

pkg/startup/autodetection/ipv4/poolselector.go
mhmxs: I can agree. It is more procedural but anyway Go is a procedural language :D

mhmxs:
@caseydavenport I fixed the code by your suggestions in a separate commit https://github.com/projectcalico/node/pull/421/commits/aa00b89f92e0fa0dd02785ae8ae2d019e2839cd5

caseydavenport:
@mhmxs sorry, but it looks like `go.mod` has a conflict. If you could rebase then we're good to merge right away. Thanks!

mhmxs:
@caseydavenport branch rebased and commits are squashed. Thanks

caseydavenport:
@mhmxs sorry to ask again... looks like another PR went in that also modified go.mod so this needs another rebase.

mhmxs:
@caseydavenport I fixed it on github. Are you able to squash and merge into one commit?

caseydavenport:
@mhmxs done, thanks much!

PR #423, opened 2020-02-14
Run checks in parallel

stamm:
## Description
Run health checks in parallel.
It works for this kind of run
```
          readinessProbe:
            exec:
              command:
              - /bin/calico-node
              - -felix-ready
              - -bird-ready
```

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

```release-note
Run health checks in parallel
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/projectcalico/felix v0.0.0-20200302175626-ffd3291c6c5a
 	github.com/projectcalico/libcalico-go v1.7.2-0.20200225165413-26809aa675f6
 	github.com/sirupsen/logrus v1.4.2
+	golang.org/x/sync v0.0.0-20190911185100-cd5d95a43a6e
 	gopkg.in/fsnotify/fsnotify.v1 v1.4.7
 
 	k8s.io/api v0.0.0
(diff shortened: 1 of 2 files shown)


caseydavenport:
@stamm could you provide some more context as to what problem this PR is trying to solve? Is there a particular bug or something that you've encountered? At a glance, it's not clear to me what the goal is, which makes it harder to review :)

stamm:
Hi. We have a problem with ready checks with calico. It run out of timeout (1s) on nodes where we have spike workflows. And we think, that it can be problem with serial checks.
But next we increase this values from default value 256 K to 1M
```
net.core.rmem_default
net.core.wmem_default
```

But I already wrote a code, and it's idiomatic for this kind of checks in go.

caseydavenport:
@stamm actually, looks like there's a conflict in go.mod. I'll give you a ping here once master is thawed and then you can do a rebase (otherwise you might get stuck rebasing a lot as go.mod changes)

caseydavenport:
@stamm I'm ready to merge this as soon as the go.mod conflict is resolved. Thanks for your patience!

stamm:
@caseydavenport Thanks, I rebase the master

PR #424, opened 2020-02-18
Add graceful restart test

neiljerram:


caseydavenport:
@neiljerram the test seems to be failing - not sure if it's a test problem or a product bug?

neiljerram:
@caseydavenport Yes, I'm not sure either. Still working on understanding that.

neiljerram:
The new test here verifies that there is no flapping of IPv4 routes.

For IPv6 routes, in the scenario of the new test, there is a flap that I have raised on the BIRD ML but is thus far unexplained: https://marc.info/?l=bird-users&m=158298182509702&w=2.  The discussion has kinda ground to a halt, but there is no indication, so far, of anything wrong in our calico/node GR setup.  I may continue trying to nudge that discussion a bit more, but personally I feel happy to close out this task, live with the known IPv6 flap, and wait until something prompts us to look at this again.

In particular, I’m confident now that our BIRD and BIRD6 are killed with -9 when the calico/node pod is deleted; in multiple runs I’ve captured and looked at the BIRD logs, and the key point is that they do not say “Shutting down” - which they would say if they were catching a TERM signal.

In terms of how that -9 happens…  I am not sure, but my best guess is:

- Kubernetes first sends a SIGTERM, which is somehow handled by runsv or our shell wrapper, and so does not reach BIRD(6) itself

- Kubernetes waits the terminationGracePeriodSeconds of 0 seconds

- Kubernetes (via the container runtime?) sends SIGKILL to all remaining processes, including BIRD(6).

PR #429, opened 2020-02-27
Move remaining build steps to semaphore v2

caseydavenport:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -4,6 +4,14 @@ agent:
   machine:
     type: e1-standard-4
     os_image: ubuntu1804
+global_job_config:
+  prologue:
+    commands:
+      - checkout
+      # Semaphore is doing shallow clone on a commit without tags.
+      # unshallow it for GIT_VERSION:=$(shell git describe --tags --dirty --always) @ Makefile.common
+      - git fetch --unshallow
+
 blocks:
 
 # Run the full set of tests. Each job can potentially be run in parallel,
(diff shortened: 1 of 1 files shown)


.semaphore/semaphore.yml
```diff
@@ -5,14 +5,50 @@ agent:
...
-          - make k8s-test
+
+# Run the full set of tests. Each job can potentially be run in parallel,
+# provided Semaphore has enough boxes available.
+- name: "Tests"
+  dependencies: []
+  task:
+    prologue:
+      commands:
+      - checkout
```
Thakurvaibhav: Since `checkout` is common for all blocks, may be use[ global_job_config](https://docs.semaphoreci.com/reference/pipeline-yaml-reference/#global_job_config)

caseydavenport: Neat, didn't spot that. Much nicer!

.semaphore/semaphore.yml
```diff
@@ -4,15 +4,47 @@ agent:
   machine:
     type: e1-standard-4
     os_image: ubuntu1804
+global_job_config:
+  prologue:
+    commands:
+      - checkout
+      # Semaphore is doing shallow clone on a commit without tags.
+      # unshallow it for GIT_VERSION:=$(shell git describe --tags --dirty --always) @ Makefile.common
+      - git fetch --unshallow
```
radTuti: perhaps consider using [semaphore cache](https://docs.semaphoreci.com/ci-cd-environment/toolbox-reference/#cache) instead of having each block checking out the same code

caseydavenport: I think I should merge this as-is to unblock CI, but that's an interesting feature - I will try to put up a follow-on PR that incorporates that. Would be nice to avoid downloading and building the code multiple times. 

maybe we could do a 
- Download, build, cache step
- Other steps un-cache the code as well as build artifacts, so they don't need even need to re-build?

PR #430, opened 2020-02-29
Don't leak old tunnel addresses

caseydavenport:
Spotted a bug where we'll leak addresses that don't have any allocation
attributes. Turns out we used to allocate all tunnel addresses without
attributes, and so tunnel addresses allocated prior to v3.6 will get
leaked once Calico is upgraded to v3.8+.

This PR aims to fix the leak. Subsequent PRs will try to clean up the
leaked addresses, and fix another potential leak spotted in code
reading.

There are a few cases we need to consider:

### The node has an address in its spec, but the address doesn't exist in IPAM.
The current code handles this already simply by assigning a new address and updating the node.

### The node has an address in its spec, the address exists in IPAM but has no attributes
This is the case that this PR is primarily trying to fix. Currently, we assign a new address and leak the old. In this PR, we will attempt to "correct" the allocation in IPAM to have attributes, and if that fails we'll just release it and assign a new one. We will only release it if we can see for sure the allocation has no handle. If it had a handle, it would suggest this is actually a WEP address.

### The node has an address in its spec, the address exists and has attributes.
The current code handles this case already

### The node has no address in its spec, the address exists in IPAM but has no attributes
In this case, the current code will simply allocate a new address and leak the old one. Since the old address has no attributes / handle, we have no way of identifying it and thus cannot clean it up.

A potential future enhancement would be to have kube-controllers detect addresses which have no handle and no attributes and garbage collect them. 

### The node has no address in its spec, the address exists in IPAM and has attributes
In this case, the current code will simply allocate a new address and leak the old one. However, the address _does_ have a handle associated with it, so we can determine if we're hitting this scenario. I propose we do a defensive "ReleaseByHandle" if we detect that the node has no address in order to cover this case. 

### The node has no address in its spec, the address does not exist in IPAM. 
The current code handles this case already. 

- [ ] Tests
- [ ] Documentation
- [ ] Release note

```release-note
Fix potential tunnel address leak when upgrading from Calico < v3.6 to Calico  >= v3.8
```


pkg/allocateip/allocateip.go
```diff
@@ -201,48 +266,50 @@ func assignHostTunnelAddr(ctx context.Context, c client.Interface, nodename stri
...
+		}
+
+		// Log the error and exit with exit code 1.
+		logCtx.WithError(err).WithField("IP", ip).Fatal("Unable to set tunnel address")
+	}
+
+	logCtx.WithField("IP", ip).Info("Set tunnel address")
+}
+
+func updateNodeWithAddress(ctx context.Context, c client.Interface, nodename string, addr string, vxlan bool) error {
```
caseydavenport: GitHub is rendering this diff a bit strangely, but all I did was pull out the logic which updates the node into a helper function so that error handling was a bit more sane.

pkg/allocateip/allocateip.go
```diff
@@ -167,11 +198,36 @@ func ensureHostTunnelAddress(ctx context.Context, c client.Interface, nodename s
...
+		Hostname: nodename,
+	}
+
+	// If we fail to allocate the same IP, return an error. We'll just
+	// have to allocate a new one. We don't need to update the node object, since
+	// it already has this address attached to it.
+	return c.IPAM().AssignIP(ctx, args)
+}
+
+func generateHandleAndAttributes(nodename string, vxlan bool) (string, map[string]string) {
```
caseydavenport: Similarly here - I extracted the logic which builds the handle and attributes so it could be resused, since we need it in a couple places now.

pkg/allocateip/allocateip_test.go
```diff
@@ -103,6 +103,109 @@ func checkTunnelAddressForNode(c client.Interface, tunnelType string, nodeName s
...
+
+	It("should not claim an address that has a handle but no attributes", func() {
+		// This test covers a scenario where the node has an IP address in its spec,
+		// but the IP has no attributes because it was assigned using an old version of Calico.
+		// In this scenario, either the allocation in IPAM has a  handle - in which case it is a
+		// WEP address - or the allocation has no handle, in which case it is a node tunnel addr.
+		// For WEP addresses, we should leave them alone and just allocate a new tunnel addr.
+
+		// Create an allocation which simulates an "old-style" allocation, prior to us
+		// using handles and attributes to attach metadata.
```
neiljerram: Comment isn't quite right because there is a handle here.

pkg/allocateip/allocateip_test.go
```diff
@@ -103,6 +103,109 @@ func checkTunnelAddressForNode(c client.Interface, tunnelType string, nodeName s
...
+		Expect(err).NotTo(HaveOccurred())
+
+		// Run the allocateip code.
+		run(nodename)
+
+		// Assert that the node no longer has the same IP on it.
+		newNode, err := c.Nodes().Get(ctx, nodename, options.GetOptions{})
+		Expect(err).NotTo(HaveOccurred())
+		Expect(newNode.Spec.BGP).NotTo(BeNil())
+		Expect(newNode.Spec.BGP.IPv4IPIPTunnelAddr).NotTo(Equal("172.16.0.1"))
```
neiljerram: But also to be another valid IP address?

pkg/allocateip/allocateip.go
```diff
@@ -134,8 +140,35 @@ func ensureHostTunnelAddress(ctx context.Context, c client.Interface, nodename s
...
+				if handle != nil {
+					// Handle exists, so this address belongs to a workload. We need to assign a new one for the node,
+					// but we shouldn't clean up the old address.
+					logCtx.WithField("currentAddr", addr).Info("Current address is occupied, assign a new one")
+				} else {
+					// Handle does not exist. This is just an old tunnel address that comes from a time before we used handles
+					// and allocation attributes. Attempt to reassign the same address, but now with metadata. It's possible
+					// that someone else takes the address while we do this, in which case we'll just need to assign a new address.
+					if err := correctAllocationWithHandle(ctx, c, addr, nodename, vxlan); err != nil {
+						if !strings.Contains(err.Error(), "already assigned in block") {
```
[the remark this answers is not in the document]
caseydavenport: Unfortunately it does not have a type, which is why I did it this way.

Maybe I should submit a libcalico PR to add a typed error instead to make this nicer.

pkg/allocateip/allocateip.go
```diff
@@ -167,11 +200,36 @@ func ensureHostTunnelAddress(ctx context.Context, c client.Interface, nodename s
...
+	args := ipam.AssignIPArgs{
+		IP:       *ipAddr,
+		HandleID: &handle,
+		Attrs:    attrs,
+		Hostname: nodename,
+	}
+
+	// If we fail to allocate the same IP, return an error. We'll just
+	// have to allocate a new one. We don't need to update the node object, since
+	// it already has this address attached to it.
```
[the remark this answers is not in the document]
caseydavenport: Yes, you're right. I think I copied this from an earlier iteration of the code, but clearly is wrong now!

pkg/allocateip/allocateip.go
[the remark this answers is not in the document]
caseydavenport: This should do it: https://github.com/projectcalico/libcalico-go/pull/1199

PR #436, opened 2020-03-11
Release tunnel addresses if the node object doesn't match

caseydavenport:
Follow on from #430 to handle the cases where the Node object has had the tunnel address in its spec either manually removed or edited to something else. This code releases any addresses that might exist if we hit either of those scenarios prior to assigning a new address. 

- [ ] Tests
- [ ] Documentation
- [ ] Release note

```release-note
None required
```

### DIFF
--- a/pkg/allocateip/allocateip.go
+++ b/pkg/allocateip/allocateip.go
@@ -118,6 +118,11 @@ func ensureHostTunnelAddress(ctx context.Context, c client.Interface, nodename s
 	if addr == "" {
 		// The tunnel has no IP address assigned, assign one.
 		logCtx.Info("Assign a new tunnel address")
+
+		// Defensively release any IP addresses with this handle. This covers a theoretical case
+		// where the node object has lost its reference to its IP, but the allocation still exists
+		// in IPAM. For example, if the node object was manually edited.
+		release = true
 	} else {
 		// Go ahead checking status of current address.
 		ipAddr := gnet.ParseIP(addr)
@@ -176,6 +181,11 @@ func ensureHostTunnelAddress(ctx context.Context, c client.Interface, nodename s
 		} else if _, ok := err.(cerrors.ErrorResourceDoesNotExist); ok {
 			// The tunnel address is not assigned, reassign it.
 			logCtx.WithField("currentAddr", addr).Info("Current address is not assigned, assign a new one")
+
+			// Defensively release any IP addresses with this handle. This covers a theoretical case
+			// where the node object has lost its reference to its correct IP, but the allocation still exists
+			// in IPAM. For example, if the node object was manually edited.
+			release = true
 		} else {
 			// Failed to get assignment attributes, datastore connection issues possible, panic
 			logCtx.WithError(err).Panicf("Failed to get assignment attributes for CIDR '%s'", addr)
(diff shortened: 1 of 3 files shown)


pkg/allocateip/allocateip.go
```diff
@@ -176,23 +181,25 @@ func ensureHostTunnelAddress(ctx context.Context, c client.Interface, nodename s
...
-		_, err := c.IPAM().ReleaseIPs(ctx, ipsToRelease)
-		if err != nil {
-			logCtx.WithField("IP", ipAddr.String()).WithError(err).Fatal("Error releasing address")
+		logCtx.WithField("IP", addr).Info("Release any old tunnel addresses")
+		handle, _ := generateHandleAndAttributes(nodename, vxlan)
+		if err := c.IPAM().ReleaseByHandle(ctx, handle); err != nil {
+			if _, ok := err.(cerrors.ErrorResourceDoesNotExist); !ok {
+				logCtx.WithError(err).Fatal("Failed to release old addresses")
+			}
+			// No existing allocations for this node.
```
neiljerram: Should we not keep the old code as well?  I.e., both release by IP (`addr`) and release by handle?

pkg/allocateip/allocateip_test.go
```diff
@@ -208,6 +208,101 @@ var _ = Describe("FV tests", func() {
 		Expect(handle).NotTo(BeNil())
 		Expect(*handle).To(Equal("some-wep-handle"))
 	})
+
+	It("should should release old IPAM addresses if they exist and the node has none", func() {
```
neiljerram: "should should"

pkg/allocateip/allocateip.go
neiljerram: (If so, should we also have a UT for the old release by IP case?)

pkg/allocateip/allocateip_test.go
```diff
@@ -208,6 +208,101 @@ var _ = Describe("FV tests", func() {
...
+		Expect(err).NotTo(HaveOccurred())
+
+		// Run the allocateip code.
+		run(nodename)
+
+		// Assert that the node no longer has the same IP on it.
+		newNode, err := c.Nodes().Get(ctx, nodename, options.GetOptions{})
+		Expect(err).NotTo(HaveOccurred())
+		Expect(newNode.Spec.BGP).NotTo(BeNil())
+		//Expect(newNode.Spec.BGP.IPv4IPIPTunnelAddr).NotTo(Equal("172.16.0.1"))
```
neiljerram: delete

pkg/allocateip/allocateip_test.go
```diff
@@ -208,6 +208,101 @@ var _ = Describe("FV tests", func() {
...
+		Expect(err).To(HaveOccurred())
+
+		// Assert that exactly one address exists for the node and that it matches the node.
+		ips, err := c.IPAM().IPsByHandle(ctx, handle)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(len(ips)).To(Equal(1))
+		Expect(newNode.Spec.BGP.IPv4IPIPTunnelAddr).To(Equal(ips[0].String()))
+	})
+
+	It("should should release old IPAM addresses if they exist and the node has a different address", func() {
```
neiljerram: "should should"

pkg/allocateip/allocateip_test.go
```diff
@@ -208,6 +208,101 @@ var _ = Describe("FV tests", func() {
...
+		Expect(err).NotTo(HaveOccurred())
+
+		// Run the allocateip code.
+		run(nodename)
+
+		// Assert that the node no longer has the same IP on it.
+		newNode, err := c.Nodes().Get(ctx, nodename, options.GetOptions{})
+		Expect(err).NotTo(HaveOccurred())
+		Expect(newNode.Spec.BGP).NotTo(BeNil())
+		Expect(newNode.Spec.BGP.IPv4IPIPTunnelAddr).To(Equal("172.16.0.0"))
```
neiljerram: Is that (172.16.0.0) reliable?

pkg/allocateip/allocateip.go
caseydavenport: I guess I was assuming that by the time we got here, either:
- The address was originally assigned with a handle, and so we're OK
- The address was originally assigned without a handle, but was corrected to have one so we're OK
- The address was originally assigned without a handle, but no longer was on the node object so we don't have an IP to use anyway.
- The address was originally assigned without a handle, but the node object was modified by the user to something wrong. I don't think its safe to just unassign whatever address was in use, since it might have been modified to match a pod IP, for example, and we wouldn't want to blindly release that address.

Does that make sense? Or am I missing a case?

pkg/allocateip/allocateip_test.go
caseydavenport: Strictly speaking, it actually is since our IPAM is deterministic given the same inputs. But we might change the code someday and that would be a confusing test failure to debug. I'll fix!

pkg/allocateip/allocateip.go
neiljerram: Just one remaining concern:

> The address was originally assigned without a handle, but was corrected to have one so we're OK

Will the correction code (in the previous PR) have run before the code here?

neiljerram: Actually, OK, I can see that we're good for all of the cases in the code above that set `release = true`.

PR #438, opened 2020-03-11
Update Felix pin for BPF MTU fix

neiljerram:


tomastigera:
BFP -> BPF

neiljerram:
Merging because the k8s-test CI failure is an unrelated problem.

PR #443, opened 2020-03-17
Release tunnel addresses by handle

caseydavenport:
## Description

This is the final PR in the trilogy for closing down IP address leaking
for tunnel addresses, and specifically covers a case where we were vulnerable to accidentally releasing an IP address that might not have been ours. 

Previous PRs in the series:
- #436 
- #430 

At this point:

- tunnel IPs will either be assigned using a handle, OR
- tunnel IPs will have been assigned without a handle and corrected to
use one, OR 
- tunnel IPs will have been assigned without a handle, and on upgrade to
the new code we determine we need to release the IP prior to correcting
it. In this case, it also will have no attributes.

Ideally we would only ever release by handle, but to cover a tiny edge case where we might not have corrected old allocations to use a handle yet, we still fallback to releasing using the exact IP address, but only if we can confirm with a lot of certainty that the IP is in fact an old address that still belongs to us (assigned prior to the use of handles and attributes).

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Release tunnel IP addresses more safely
```

### DIFF
--- a/pkg/allocateip/allocateip.go
+++ b/pkg/allocateip/allocateip.go
@@ -361,9 +361,30 @@ func removeHostTunnelAddr(ctx context.Context, c client.Interface, nodename stri
 			node.Spec.BGP = nil
 		}
 
-		// Release the IP.
-		if _, err := c.IPAM().ReleaseIPs(ctx, []net.IP{*ipAddr}); err != nil {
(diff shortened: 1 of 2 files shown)


caseydavenport:
@neiljerram thanks, yeah I reworked the PR contents after I put it up. Originally the PR did actually always release using handle :) 

I've updated the title and description.

neiljerram:
Lovely, thanks.

PR #448, opened 2020-03-23
Include BIRD source in docker image

caseydavenport:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -111,6 +111,7 @@ endif
 # Versions and location of dependencies used in the build.
 BIRD_VER?=v0.3.3-0-g1e8dd375
 BIRD_IMAGE ?= calico/bird:$(BIRD_VER)-$(ARCH)
+BIRD_SOURCE=filesystem/included-source/bird-$(BIRD_VER).tar.gz
 
 # Versions and locations of dependencies used in tests.
 CALICOCTL_VER?=master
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -167,6 +167,7 @@ clean:
 	find . -name '*.created' -exec rm -f {} +
 	find . -name '*.pyc' -exec rm -f {} +
 	rm -rf certs *.tar vendor $(NODE_CONTAINER_BIN_DIR)
+	rm -rf filesystem/included-source/*
```
neiljerram: Is that better in some way than `rm -rf filesystem/included-source` ?  I find the `*` a little uncomfortable there, although I don't think it introduces any real problem.

Makefile
```diff
@@ -267,6 +268,12 @@ endif
 	docker build --pull -t $(BUILD_IMAGE):latest-$(ARCH) . --build-arg BIRD_IMAGE=$(BIRD_IMAGE) --build-arg QEMU_IMAGE=$(CALICO_BUILD) --build-arg ver=$(CALICO_GIT_VER) -f ./Dockerfile.$(ARCH)
 	touch $@
 
+# download BIRD source to include in image.
+bird-source: filesystem/included-source/bird-$(BIRD_VER).tar.gz
```
neiljerram: Avoid a phony target by saying instead
`BIRD_SOURCE = filesystem/included-source/bird-$(BIRD_VER).tar.gz`
and then using `$(BIRD_SOURCE)` above instead of `bird-source`.

caseydavenport:
@rafaelvanoni @neiljerram thanks for the reviews. I think I've addressed feedback now.

PR #457, opened 2020-03-31
[phone home] Distinguish Rancher deployments

rafaelvanoni:


pkg/startup/startup.go
```diff
@@ -1025,6 +1040,14 @@ func ensureDefaultConfig(ctx context.Context, cfg *apiconfig.CalicoAPIConfig, c
 		}
 	}
 
+	if rancherState != nil {
```
tmjd, line 1043: I know we didn't do this before but should we be checking the contents of the clusterType to make sure rancher isn't already part of the clusterType?
We could have a common function that could be used to check if it is already present and add appropriately if not, then that could be used here and for kubeadm.

caseydavenport: I think this de-duplication is already performed in the CRUD logic for clusterinformations: https://github.com/projectcalico/libcalico-go/blob/9afaf487397901c920a6d6f620c8e8d308ee1dd5/lib/clientv3/client.go#L295

Issue #463, opened 2020-04-16
Return the true error message when socket path is /var/run/calico but BIRD service is exception


marviniter:
When connecting to the BIRD socket in `/var/run/calico` to get the data,if that fails and the error is "connect: no such file or directory", then we can try connecting to BIRD socket in `/var/run/bird`. but when the error is others,such as "connect: resource temporarily unavailable",we should return the true error message.
## Expected Behavior

```
# ls /var/run/calico
bird.ctl   bird6.ctl
/ # ls /var/run/bird
ls: /var/run/bird: No such file or directory
```
the output error message is "Error querying BIRD: unable to connect to BIRDv4 socket:dial unix /var/run/bird/bird.ctl: connect: no such file or directory",but the true error message is "connect: resource temporarily unavailable"

## Current Behavior

when benchmark the BIRD service and execute `calico-node -bird-ready` command,the error message is "unable to connect to BIRDv4 socket: dial unix /var/run/bird/bird.ctl: connect: no such file or directory" 

## Possible Solution

if the error is "connect: no such file or directory",then  try connecting to BIRD socket in `/var/run/bird`,otherise return the error message direct

## Your Environment

* calico/node version:  v3.3.1
* Operating System and version: CentOS Linux release 7.6.1810 
* Link to your project (optional):

marviniter:
Have merged, closing

PR #464, opened 2020-04-16
Fix bird health check return error message when BIRD service report "connect: resource temporarily unavailable"

marviniter:
## Description
bug fix, add error type judge before retry connect to `/var/run/bird`
Closes:https://github.com/projectcalico/node/issues/463

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
When connecting to BIRD  check the BIRD socket file
```

### DIFF
--- a/pkg/health/bird/bird.go
+++ b/pkg/health/bird/bird.go
@@ -39,6 +39,10 @@ func GRInProgress(ipv string) (bool, error) {
 	// Try connecting to the BIRD socket in `/var/run/calico/` first to get the data
 	c, err := net.Dial("unix", fmt.Sprintf("/var/run/calico/bird%s.ctl", birdSuffix))
 	if err != nil {
+		// If that fails and BIRD socket file exists, return the error message
+		if !isNotExistError(err) {
+			return false, fmt.Errorf("Error querying BIRD: connect to BIRDv%s socket: %v", ipv, err)
+		}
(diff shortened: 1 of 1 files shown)


pkg/health/bird/bird.go
```diff
@@ -39,6 +39,10 @@ func GRInProgress(ipv string) (bool, error) {
 	// Try connecting to the BIRD socket in `/var/run/calico/` first to get the data
 	c, err := net.Dial("unix", fmt.Sprintf("/var/run/calico/bird%s.ctl", birdSuffix))
 	if err != nil {
+		// If that fails and BIRD socket file exists, return the error message
+		if !isNotExistError(err) {
```
fasaxc, line 43: does `os.IsNotExist()` work for the errors returned by `net.Dial`?

marviniter: Thanks,i will use os.IsNotExist to replace the checking error text.

tmjd:
@marviniter would you mind updating the "Release Note" in the description.
@fasaxc I think this is ready for another review.

marviniter:
> @marviniter would you mind updating the "Release Note" in the description.
> @fasaxc I think this is ready for another review.

Thanks,i've updated!

marviniter:
> Definitely looks like an improvement.

Thanks!

PR #474, opened 2020-04-23
Fix tests and update pins

lmm:
## Description

This PR includes a pin update for libcalico-go and a go-build version bump.

The code bits of this PR fix the python STs which are failing now because of the new projectcalico-default-allow profile that always exists. The tests expect that it is possible to apply the saved profiles to restore them, which is not the case for projectcalico-default-allow (you can't update this profile).

Example output from a failed run:
```
======================================================================
ERROR: test_rules_dest_ip_nets (tests.st.policy.test_profile.MultiHostMainline)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/code/tests/st/policy/test_profile.py", line 273, in test_rules_dest_ip_nets
    self._apply_new_profile(self.new_profiles, self.host1)
  File "/code/tests/st/policy/test_profile.py", line 397, in _apply_new_profile
    host.calicoctl("apply -f new_profiles")
  File "/code/tests/st/utils/docker_host.py", line 306, in calicoctl
    return self.execute(calicoctl + " " + command, raise_exception_on_failure=raise_exception_on_failure)
  File "/code/tests/st/utils/docker_host.py", line 223, in execute
    return log_and_run(command, raise_exception_on_failure=raise_exception_on_failure)
  File "/code/tests/st/utils/utils.py", line 118, in log_and_run
    raise CommandExecError(e)

CommandExecError: Command docker exec -it host1 sh -c 'export ETCD_ENDPOINTS=http://192.168.122.201:2379; export ETCD_CA_CERT_FILE=; export ETCD_CERT_FILE=; export ETCD_KEY_FILE=; /code/dist/calicoctl apply -f new_profiles' failed with RC 1 and output:

Partial success: applied the first 2 out of 3 'Profile' resources:

Hit error(s): [operation Update is not supported on projectcalico-default-allow: The profile "projectcalico-default-allow" is a default provided by Calico and cannot be updated]
```

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


tests/st/policy/test_profile.py
```diff
@@ -60,6 +60,15 @@ def setUp(self):
...
         output = host1.calicoctl("get profile -o yaml")
         self.original_profiles = yaml.safe_load(output)['items']
+
+        # Filter out the default-allow profile
+        temp = []
+        for profile in self.original_profiles:
+            if profile['metadata']['name'] != "projectcalico-default-allow":
+                temp.append(profile)
+                    
+        self.original_profiles = temp
```
lmm, line 70: This bit of code saves copy of the profiles before tests executed. That original copy of the profiles is restored after a test. In this case, we just remove the default-allow profile from the original list.

Issue #486, opened 2020-05-05
Respect FELIX_HEALTHHOST in felixReadinessEp


gdziwoki:
Health check and readiness check respect environment variable FELIX_HEALTHPORT but not FELIX_HEALTHHOST.
https://github.com/projectcalico/node/blob/28ef5a6a93a3330da2067cfdc406286946028e13/pkg/health/health.go#L48

## Expected Behavior
`/bin/calico-node -felix-health` should respect both env vars

## Current Behavior
If we change FELIX_HEALTHHOST to something different than localhost, the FELIX_HEALTHHOST:FELIX_HEALTHPORT/liveness endpoint will be available on that address, but `/bin/calico-node -felix-health` will fail

## Your Environment
* calico/node version: 3.13.3
* Operating System and version: Using helm chart

tmjd:
Thank you for submitting this and the PR to fix it.
Could you share your use case for needing to set the HEALTHHOST for felix? I'm wondering if instead of adding this if we should just be removing from the documentation the ability to set the healthhost.

gdziwoki:
We want to monitor calico-node using node-problem-detector, which means accessing it through nodes' IP address.
Right now, with default settings the built-in,local healthcheck will work, but the one performed by node-problem-detector will not.
If I change FELIX_HEALTHHOST, the external test will start working, but the built in healthcheck will fail.

PR #487, opened 2020-05-05
Added conditional check for FELIX_HEALTHHOST env variable

bartek-lopatka:
## Description
Link to the issue: https://github.com/projectcalico/node/issues/486

Just added check for FELIX_HEALTHHOST env variable

### DIFF
--- a/pkg/health/health.go
+++ b/pkg/health/health.go
@@ -35,25 +35,17 @@ var felixLivenessEp string
 
 func init() {
 	felixPort := os.Getenv("FELIX_HEALTHPORT")
-	felixHost := os.Getenv("FELIX_HEALTHHOST")
-	if felixPort == "" && felixHost == "" {
-		felixReadinessEp = "http://localhost:9099/readiness"
-		felixLivenessEp = "http://localhost:9099/liveness"
-		return
-
-	} else if felixPort == "" {
-		felixReadinessEp =  "http://" + felixHost + ":9099/readiness"
-		felixLivenessEp = "http://" + felixHost + ":9099/liveness"
-		return
-
-	} else if felixHost == "" {
-		felixReadinessEp = "http://localhost:" + felixPort + "/readiness"
-		felixLivenessEp = "http://localhost:" + felixPort + "/liveness"
-		return
+	if felixPort == "" {
+		felixPort = "9099"
 	}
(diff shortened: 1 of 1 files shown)


pkg/health/health.go
```diff
@@ -35,18 +35,29 @@ var felixLivenessEp string
...
+	} else if felixPort == "" {
+		felixReadinessEp =  "http://" + felixHost + ":9099/readiness"
+		felixLivenessEp = "http://" + felixHost + ":9099/liveness"
+		return
+
+	} else if felixHost == "" {
+		felixReadinessEp = "http://localhost:" + felixPort + "/readiness"
+		felixLivenessEp = "http://localhost:" + felixPort + "/liveness"
+		return
 	}
```
tmjd, line 53: Thank you for submitting this.
What do you think about changing this a bit?
I think instead of dealing with the 4 different cases (L40, L45, L50, and L59) it would be good to do something like
```
felixPort := os.Getenv("FELIX_HEATHPORT")
if felixPort == "" {
  felixPort = "9099"
}
if _, err := strconv.Atoi(felixPort); err != nil {
￼  log.Panicf("Failed to parse value for port %q", felixPort)
￼}

felixHost := os.Getenv("FELIX_HEALTHHOST")
if felixHost == "" {
  felixHost = localhost
}
felixReadinessEp = "http://" + felixHost + ":" + felixPort + "/readiness"
felixLivenessEp = "http://" + felixHost + ":" + felixPort + "/liveness"
```

bartek-lopatka: Yeah, your proposition makes more sense. Will change that in a moment.

caseydavenport:
@bartek-lopatka thanks! Could you please also submit a PR to update the documentation at github.com/projectcalico/calico for this change?

bartek-lopatka:
@caseydavenport sure, but how exactly? I've found instructions (https://github.com/projectcalico/calico/blob/master/CONTRIBUTING_DOCS.md) but it's not clear to me what files I should modify.

caseydavenport:
@bartek-lopatka sorry for missing your reply. Just took a look at this again and I think you're all good, actually. Existing documentation should cover the changes here :+1: 

Thanks again for your contribution!

PR #508, opened 2020-06-01
Use a tunnel IP allocation daemon rather than using confd to trigger IP allocations

robbrockbank:
## Description
Updates the tunnel ip allocation script to work in two modes:  single shot and long running daemon.

The daemon uses a syncer (direct or via typha) to watch IP pools and node status information and uses that to determine if the IP addresses need to be reconciled.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/allocateip/allocateip.go
+++ b/pkg/allocateip/allocateip.go
@@ -8,79 +8,228 @@ import (
 	"reflect"
 	"time"
 
+	log "github.com/sirupsen/logrus"
+
+	"github.com/projectcalico/libcalico-go/lib/apiconfig"
 	api "github.com/projectcalico/libcalico-go/lib/apis/v3"
 	v3 "github.com/projectcalico/libcalico-go/lib/apis/v3"
+	bapi "github.com/projectcalico/libcalico-go/lib/backend/api"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+	"github.com/projectcalico/libcalico-go/lib/backend/syncersv1/tunnelipsyncer"
 	client "github.com/projectcalico/libcalico-go/lib/clientv3"
 	cerrors "github.com/projectcalico/libcalico-go/lib/errors"
 	"github.com/projectcalico/libcalico-go/lib/ipam"
 	"github.com/projectcalico/libcalico-go/lib/net"
 	"github.com/projectcalico/libcalico-go/lib/options"
+	"github.com/projectcalico/typha/pkg/syncclientutils"
+	"github.com/projectcalico/typha/pkg/syncproto"
+
+	"github.com/projectcalico/node/buildinfo"
 	"github.com/projectcalico/node/pkg/calicoclient"
-	"github.com/sirupsen/logrus"
-	log "github.com/sirupsen/logrus"
 )
 
 // This file contains the main processing and common logic for assigning tunnel addresses,
-// used by calico/node to set the host's tunnel address if IPIP or VXLAN is enabled.
-// It will assign an address address if there are any available, and remove any tunnel address
-// that is configured if it should no longer be.
+// used by calico/node to set the host's tunnel address if IPIP or VXLAN is enabled on the pool or
+// wireguard is enabled on the node.
+//
+// It will assign an address if there are any available, and remove any tunnel addresses
+// that are configured and should no longer be.
 
-func Run() {
+// Run runs the tunnel ip allocator. If done is nil, it runs in single-shot mode. If non-nil, it runs in daemon mode
+// performing a reconciliation when IP pool or node configuration changes that may impact the allocations.
+func Run(done <-chan struct{}) {
 	// This binary is only ever invoked _after_ the
 	// startup binary has been invoked and the modified environments have
 	// been sourced.  Therefore, the NODENAME environment will always be
 	// set at this point.
 	nodename := os.Getenv("NODENAME")
 	if nodename == "" {
-		logrus.Panic("NODENAME environment is not set")
+		log.Panic("NODENAME environment is not set")
 	}
-
-	run(nodename)
-}
-
-func run(nodename string) {
 
 	// Load the client config from environment.
 	cfg, c := calicoclient.CreateClient()
 
-	// If configured to use host-local IPAM, this script has nothing to do, so just return.
-	if cfg.Spec.K8sUsePodCIDR {
-		logrus.Debug("Using host-local IPAM, no need to allocate a tunnel IP")
+	run(nodename, cfg, c, done)
+}
+
+func run(nodename string, cfg *apiconfig.CalicoAPIConfig, c client.Interface, done <-chan struct{}) {
+	if done == nil {
+		// Running in single shot mode, so assign addresses and exit.
+		reconcileTunnelAddrs(nodename, cfg, c)
 		return
 	}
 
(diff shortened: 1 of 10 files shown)


pkg/allocateip/allocateip.go
```diff
@@ -8,79 +8,228 @@ import (
 	"reflect"
 	"time"
 
+	log "github.com/sirupsen/logrus"
```
caseydavenport, line 11: I think we've been trying to avoid aliasing `logrus` to `log` elsewhere. Do we need to do this?

caseydavenport: I guess it's fine. I see we were importing it as both before. But I would have removed the aliased one rather than the other way around /shrug

robbrockbank: I've no issue switching it up - didn't realize we were avoiding log.

pkg/allocateip/allocateip.go
```diff
@@ -350,8 +528,14 @@ func removeHostTunnelAddr(ctx context.Context, c client.Interface, nodename stri
 			node.Spec.BGP = nil
 		}
 
+		// Sim. for wireguard.
```
caseydavenport, line 531: Sim?

pkg/allocateip/allocateip.go
```diff
@@ -326,21 +487,38 @@ func removeHostTunnelAddr(ctx context.Context, c client.Interface, nodename stri
...
+				node.Spec.BGP.IPv4IPIPTunnelAddr = ""
+			}
+		case ipam.AttributeTypeWireguard:
+			if node.Spec.Wireguard != nil {
+				ipAddr = net.ParseIP(node.Spec.Wireguard.InterfaceIPv4Address)
+				node.Spec.Wireguard.InterfaceIPv4Address = ""
+			}
+		}
+
+		// If no IP address found then there is nothing to remove.
```
caseydavenport, line 518: I don't think this is correct - it's possible that the node and IPAM data is out of sync. We should attempt to remove the address every time no matter what rather than short circuit here, otherwise we might leak IP addresses. 

We release by handle so that we don't need the address in hand in order to do the release.

caseydavenport: Hrm, maybe that's not true in this code path. Seems like we assume the IP is non-nil below...

robbrockbank: Hmmm, not sure why I added that - certainly didn't mean to change the existing behavior.  My bad!

pkg/allocateip/allocateip.go
caseydavenport: We're just not consistent with it. Sometimes we alias, sometimes we declare a logger instance named `log`. It's just all over the place, so I wouldn't worry about it..

pkg/allocateip/allocateip.go
caseydavenport: I think that is a bug, though. Even if there is no IP on the node, we should try to release it by handle anyway to clean up anything that might be left over in IPAM.

pkg/allocateip/allocateip.go
```diff
@@ -326,21 +487,38 @@ func removeHostTunnelAddr(ctx context.Context, c client.Interface, nodename stri
 		}
 
 		// Determine if we need to do any work.
-		ipipTunnelAddrExists := (node.Spec.BGP != nil && node.Spec.BGP.IPv4IPIPTunnelAddr != "")
+		ipipTunnelAddrExists := node.Spec.BGP != nil && node.Spec.BGP.IPv4IPIPTunnelAddr != ""
 		vxlanTunnelAddrExists := node.Spec.IPv4VXLANTunnelAddr != ""
-		if (vxlan && !vxlanTunnelAddrExists) || (!vxlan && !ipipTunnelAddrExists) {
```
robbrockbank, line 331: @caseydavenport  - this bit of code would have ensured the IP address was assigned a value.  So I figured that was the intent.

My additional check below really is a back up in case it doesn't parse as an IP - but I don't think I've changed the intent.

pkg/allocateip/allocateip.go
```diff
@@ -399,7 +583,14 @@ func removeHostTunnelAddr(ctx context.Context, c client.Interface, nodename stri
 
 // determineEnabledPools returns all enabled pools. If vxlan is true, then it will only return VXLAN pools. Otherwise
 // it will only return IPIP enabled pools.
-func determineEnabledPoolCIDRs(node api.Node, ipPoolList api.IPPoolList, vxlan bool) []net.IPNet {
+func determineEnabledPoolCIDRs(node api.Node, ipPoolList api.IPPoolList, attrType string) []net.IPNet {
+	// For wireguard, return no valid pools if the wireguard public key has not been set. Only once wireguard has been
+	// enabled *and* the wireguard device has been initialized do we require an IP address to be configured.
```
caseydavenport, line 588: Don't we need the IP in order to init the wg interface?

I think we'll end up with:
- Felix creates interface, writes back public key
- This code spots it, assigns IP
- Felix spots the IP, and writes it to the wg interface.

Is that right? It will work, it's just different from how we do VXLAN IIUC, since for VXLAN we don't create the interface until the IP has been allocated, since the overlay won't work properly until the IP is programmed on it.

pkg/allocateip/allocateip.go
robbrockbank: But then take your point that there was probably a bug there.

pkg/allocateip/allocateip.go
robbrockbank: That is correct.  We can configure wg without an IP and then update it after.  I think this is OK because it's lightweight - in that felix doesn't do a full restart because the interface address was updated.

pkg/allocateip/allocateip.go
```diff
@@ -416,16 +607,23 @@ func determineEnabledPoolCIDRs(node api.Node, ipPoolList api.IPPoolList, vxlan b
...
 				cidrs = append(cidrs, *poolCidr)
 			}
-		} else {
+		case ipam.AttributeTypeIPIP:
 			// Check if IPIP is enabled in the IP pool, the IP pool is not disabled, and it is IPv4 pool since we don't support IPIP with IPv6.
 			if (ipPool.Spec.IPIPMode == api.IPIPModeCrossSubnet || ipPool.Spec.IPIPMode == api.IPIPModeAlways) && !ipPool.Spec.Disabled && poolCidr.Version() == 4 {
 				cidrs = append(cidrs, *poolCidr)
 			}
+		case ipam.AttributeTypeWireguard:
+			// Wireguard does not require a specific encap configuration on the pool.
```
caseydavenport, line 623: I think we should respect the IP pool's nodeSelector field though, right? I think you should double check that the IPAM code handles that properly when explicit IP pools are passed in. Otherwise we might need to do a selector match here.

robbrockbank: Yes we should - the node selector check is done above ^^^ so I think this is correct as it stands.

That said, it's also done in the IPAM code, so I could simplify this logic and remove the pool selection code entirely (since IPAM checks a pool is enabled and matches the node selector) and just return nil pools to allow the IPAM code to select pools for itself.  Having the logic here though seems less confusing.

Issue #512, opened 2020-06-03
Calico node doesn't support disabling ip autodetection


krisiasty:
## Expected Behavior

According to documentation (https://docs.projectcalico.org/reference/node/configuration#ip-setting) if IP environment variable is not set or is set to empty string, calico should not perform autodetection and use previously configured ip address. When using tigera-operator to deploy calico and configuring nodeAddressAutodetectionV4 property to empty object {}, the IP environment variable in generated daemon set is configured to value "none" and calico node should respect this setting.

## Current Behavior

tigera-operator can only set the IP environment variable to value "none" or "autodetect" with IP_AUTODETECTION_METHOD  set to one of the "interface=", "skip-interface=", "can-reach=", or "first-find=", as seen in the following code:
https://github.com/tigera/operator/blob/ee329892ca56bed6903037fa1b8069f4dae98e0b/pkg/render/node.go#L817
https://github.com/tigera/operator/blob/ee329892ca56bed6903037fa1b8069f4dae98e0b/pkg/render/node.go#L1030

Yet the following code in calico node doesn't check for previously configured IP address and terminates as soon as it sees IP="none" and (IP6="none" or IP6=""), which is clearly wrong and not what documentation says:
https://github.com/projectcalico/node/blob/58a1b6a4a9cde11e6047e78a4550c529f5f468fe/pkg/startup/startup.go#L488-L491

## Possible Solution

The code mentioned above should also check for previously configured IP and terminate only if there is none present.

## Steps to Reproduce (for bugs)

1. deploy calico using tigera-operator
2. modify custom resource installations.operator.tigera.io/default and change:

>   nodeAddressAutodetectionV4:
>       firstFound: true

to:

> nodeAddressAutodetectionV4: {}

3. confirm that environment variable IP in the calico-node daemon set is changed to:

>     - name: IP
>        value: none

4. deamon set should start rolling upgrade to apply the change to the pods
5. first pod should end in the CrashLoopBackOff state with the log stating:

> 2020-06-03 19:44:37.519 [INFO][9] startup.go 309: Early log level set to info
> 2020-06-03 19:44:37.519 [INFO][9] startup.go 325: Using NODENAME environment for node name
> 2020-06-03 19:44:37.519 [INFO][9] startup.go 337: Determined node name: master1
> 2020-06-03 19:44:37.521 [INFO][9] startup.go 369: Checking datastore connection
> 2020-06-03 19:44:37.529 [INFO][9] startup.go 393: Datastore connection verified
[...]

PR #513, opened 2020-06-04
Properly handle disabled autodetection

krisiasty:
## Description

If IP or IP6 environment variable is set to "none" try to use existing address and not terminate. This fixes calico to work as specified in documentation.

Fixes https://github.com/projectcalico/node/issues/512

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix that calico/node required IP auto detection to be enabled
```

### DIFF
--- a/pkg/startup/startup.go
+++ b/pkg/startup/startup.go
@@ -453,6 +453,9 @@ func configureIPsAndSubnets(node *api.Node) (bool, error) {
 			log.Warnf("Autodetection of IPv4 address failed, keeping existing value: %s", node.Spec.BGP.IPv4Address)
 			validateIP(node.Spec.BGP.IPv4Address)
 		}
+	} else if ipv4Env == "none" && node.Spec.BGP.IPv4Address != "" {
+		log.Warnf("Autodetection for IPv4 disabled, keeping existing value: %s", node.Spec.BGP.IPv4Address)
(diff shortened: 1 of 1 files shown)


pkg/startup/startup.go
```diff
@@ -453,6 +453,9 @@ func configureIPsAndSubnets(node *api.Node) (bool, error) {
 			log.Warnf("Autodetection of IPv4 address failed, keeping existing value: %s", node.Spec.BGP.IPv4Address)
 			validateIP(node.Spec.BGP.IPv4Address)
 		}
+	} else if ipv4Env == "none" && node.Spec.BGP.IPv4Address != "" {
+		log.Warnf("Autodetection for IPv4 disabled, keeping existing value: %s", node.Spec.BGP.IPv4Address)
```
caseydavenport, line 457: I don't think this needs to be a warning. Should probably be an "Info" level log (since this is an expected code path).

caseydavenport:
@krisiasty thanks for this PR! I had spotted this a little while ago but hadn't gotten around to fixing.

I think the fix looks good - I just had one comment on the log level.

/sem-approve

pkg/startup/startup.go
krisiasty: Changed to log.Infof as suggested. Same for ipv6

PR #518, opened 2020-06-12
Introduce detection method cidr

mandelsoft:
## Description
New Feature: Offer a new detection method for a node's primary IP and interface.

The current implementation typically fails to identify the correct primary interface, if there is a tun/tap device already present if the lookup is done. It then finds the tun device and the rest of the node setup is done incorrectly.

So far, there is only one way to avoid this: use of the detection method `interface`, but it requires to know the interface name, which is typically operating system specific. To avoid this in a more generic setup as it is done for example in the [gardener](https://github.com/gardener/gardener) scenario it would be helpful to be able to identify the interface and IP address by a node IP range, that is typically given in an operating system independent manner and already known, when the nodes of a Kubernetes cluster are created.

Therefore this PR implements the detection method `cidr`, which allows to specify IP ranges (given by CIDRs). The first interface is selected, which matches one of the given ranges.

The already existing function `FilteredEnumeration` is just extended by an optional additional filter condition using CIDRs and the `Contains` method of the `net.IPNet` structure.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
A new IP/interface detection method `cidr` is added. The syntax (for example for the environment variable `IP_AUTODETECTION_METHOD` is `cidr=<cidr>(,<cidr>)*`.
```

### DIFF
--- a/pkg/startup/startup.go
+++ b/pkg/startup/startup.go
@@ -38,9 +38,6 @@ import (
 	"github.com/projectcalico/libcalico-go/lib/selector"
 	"github.com/projectcalico/libcalico-go/lib/upgrade/migrator"
 	"github.com/projectcalico/libcalico-go/lib/upgrade/migrator/clients"
-	"github.com/projectcalico/node/pkg/calicoclient"
-	"github.com/projectcalico/node/pkg/startup/autodetection"
-	"github.com/projectcalico/node/pkg/startup/autodetection/ipv4"
 	log "github.com/sirupsen/logrus"
 	kapiv1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
@@ -48,6 +45,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
+
+	"github.com/projectcalico/node/pkg/calicoclient"
+	"github.com/projectcalico/node/pkg/startup/autodetection"
+	"github.com/projectcalico/node/pkg/startup/autodetection/ipv4"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


pkg/startup/startup.go
```diff
@@ -630,6 +632,15 @@ func autoDetectCIDR(method string, version int) *cnet.IPNet {
 		// Regexes are passed in as a string separated by ","
 		ifRegexes := regexp.MustCompile(`\s*,\s*`).Split(ifStr, -1)
 		return autoDetectCIDRByInterface(ifRegexes, version)
+	} else if strings.HasPrefix(method, AUTODETECTION_METHOD_CIDR) {
+		// Autodetect the IP by filtering interface by its address.
+		cidrStr := strings.TrimPrefix(method, AUTODETECTION_METHOD_CIDR)
+		// CIDRs are passed in as a string separated by ","
+		matches:=[]cnet.IPNet{}
+		for _, r:= range regexp.MustCompile(`\s*,\s*`).Split(cidrStr, -1) {
+			matches=append(matches, cnet.MustParseCIDR(r))
```
caseydavenport, line 641: We shouldn't use `MustParseCIDR` here - that will cause a panic if the input is not a valid CIDR.

Instead, we should parse and handle the error gracefully.

pkg/startup/startup.go
```diff
@@ -60,6 +61,7 @@ const (
 	AUTODETECTION_METHOD_CAN_REACH      = "can-reach="
 	AUTODETECTION_METHOD_INTERFACE      = "interface="
 	AUTODETECTION_METHOD_SKIP_INTERFACE = "skip-interface="
+	AUTODETECTION_METHOD_CIDR           = "cidr="
```
caseydavenport, line 64: I think this should be called `cidrs` since it's a list

caseydavenport: I guess our other options here are singular, even though they support multiple options. So you can ignore this comment.

pkg/startup/startup.go
mandelsoft: fixed

caseydavenport:
@mandelsoft looks like you've got a build failure:

```

github.com/projectcalico/node/pkg/startup/autodetection_test [github.com/projectcalico/node/pkg/startup/autodetection.test]
--
1269 | pkg/startup/autodetection/filtered_test.go:28:57: not enough arguments in call to autodetection.FilteredEnumeration
1270 | have (nil, nil, number)
1271 | want ([]string, []string, []"github.com/projectcalico/libcalico-go/lib/net".IPNet, int)

```

mandelsoft:
Sorry about the delay, I was completely busy last week.

mandelsoft:
Tests are not executed. Do I still have to do something?

caseydavenport:
@mandelsoft wow sorry, missed that I hadn't triggered tests. They should run now.

PR #536, opened 2020-07-29
Build as root if UID 0

caseydavenport:
## Description

In environments where the build must be run as root (e.g., some CI systems run as root), our build fails because this flag isn't set.

Detect these scenarios and set the flag, which triggers this: https://github.com/projectcalico/go-build/blob/v0.45/entrypoint.sh#L9 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 PACKAGE_NAME?=github.com/projectcalico/node
-GO_BUILD_VER?=v0.40
+GO_BUILD_VER?=v0.45
 
 # This needs to be evaluated before the common makefile is included.
 # This var contains some default values that the common makefile may append to.


Makefile
```diff
@@ -40,6 +40,11 @@ endif
 
 EXTRA_DOCKER_ARGS+=-e GOPRIVATE='github.com/tigera/*'
 
+ifeq ("$(LOCAL_USER_ID)", "0")
+# The build needs to run as root.
+EXTRA_DOCKER_ARGS+=-e RUN_AS_ROOT='true'
+endif
+
```
realgaurav, line 47: Can this be moved to Makefile.common?

caseydavenport: yeah, that's probably a better place.

caseydavenport: https://github.com/projectcalico/go-build/pull/122

PR #541, opened 2020-08-04
Don't fail if not authorized to access configmaps

caseydavenport:
## Description

We query kubeadm / rancher configmaps to get some cluster info. However,
for platforms that aren't those, we don't need to give permissions to
access those configmaps. 

This handles that gracefully.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/startup/startup.go
+++ b/pkg/startup/startup.go
@@ -139,11 +139,13 @@ func Run() {
 
 		// Check if we're running on a kubeadm and/or rancher cluster. Any error other than not finding the respective
 		// config map should be serious enough that we ought to stop here and return.
-		kubeadmConfig, err = clientset.CoreV1().ConfigMaps(metav1.NamespaceSystem).Get(KubeadmConfigConfigMap,
-			metav1.GetOptions{})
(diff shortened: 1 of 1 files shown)


pkg/startup/startup.go
```diff
@@ -139,11 +139,13 @@ func Run() {
 
 		// Check if we're running on a kubeadm and/or rancher cluster. Any error other than not finding the respective
 		// config map should be serious enough that we ought to stop here and return.
-		kubeadmConfig, err = clientset.CoreV1().ConfigMaps(metav1.NamespaceSystem).Get(KubeadmConfigConfigMap,
-			metav1.GetOptions{})
+		kubeadmConfig, err = clientset.CoreV1().ConfigMaps(metav1.NamespaceSystem).Get(KubeadmConfigConfigMap, metav1.GetOptions{})
 		if err != nil {
 			if errors.IsNotFound(err) {
 				kubeadmConfig = nil
+			} else if errors.IsUnauthorized(err) {
```
bcreane, line 146: this looks like a reasonable inference and an improvement to spamming the log with errors that are not relevant.

fasaxc:
I like the approach; log message is clear and it's better than falling on our face.

Issue #524, opened 2020-06-22
Calico/node aarch64 image iptables version is different from amd64 images


LyleLee:
Calico/node aarh64 image iptables version is different from amd64 images. 
## Expected Behavior

amd64 image calico/node:v3.14.1 should have iptables v1.8.2 (legacy)
amd64 image calico/node:latest should have iptables v1.8.2 (legacy)
aarch64 image calico/node:v3.14.1 should have iptables v1.8.2 (legacy)
aarch64 image calico/node:latest should have iptables v1.8.2 (legacy)
## Current Behavior

```
[root@master1 ~]# uname -a
Linux master1 4.18.0-147.el8.x86_64 #1 SMP Wed Dec 4 21:51:45 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
[root@master1 ~]# docker run --rm -it --name test calico/node:v3.14.1 iptables --version
iptables v1.8.2 (legacy)
[root@master1 ~]# docker run --rm -it --name test calico/node:latest iptables --version
iptables v1.8.2 (legacy)
[root@master1 ~]#

```
But we got so much different on aarh64 platform:

```
[root@master1 ~]# uname -a
Linux master1 4.18.0-147.el8.aarch64 #1 SMP Wed Dec 4 21:57:21 UTC 2019 aarch64 aarch64 aarch64 GNU/Linux
[root@master1 ~]# docker run --rm -it --name test calico/node:v3.14.1 iptables --version
iptables v1.6.2
[root@master1 ~]# docker run --rm -it --name test calico/node:latest iptables --version
iptables v1.6.2
[root@master1 ~]#

```

## Possible Solution

Use the same multi-arch suported image base at Dockerfile, and pack iptables iptables v1.8.2 (legacy). 

amd64 base image now is : [CentOS8](https://github.com/projectcalico/node/blob/355d597e9bffd72a93546c078646d13712d85b11/Dockerfile.amd64#L27)

aarch64 base image now is: [arm64v8/alpine](https://github.com/projectcalico/node/blob/355d597e9bffd72a93546c078646d13712d85b11/Dockerfile.arm64#L22)

Can we all switch CentOS8?

## Addition 

Actually I have the problem described here https://github.com/projectcalico/calico/issues/2322

My pods on Kubernetes can not ping through each other.

[...]

PR #543, opened 2020-08-12
Dockerfile.arm64 Catch up Dockerfile.amd64

LyleLee:
Adopt the same iptable version

With CentOS8 or latest debian OS, Whose iptables implementation
have change from Xtables to nftables. Need to pack legacy
iptables into node image.

## Description

+ Calico/node aarh64 image iptables version is different from amd64 images.
+ Iptables rules not writing into host.
+ Ping between pod not working.

Try to fix issues:
https://github.com/projectcalico/node/issues/524

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
None required
```


Makefile
```diff
@@ -53,8 +54,8 @@ FELIX_GPL_SOURCE=filesystem/included-source/felix-ebpf-gpl.tar.gz
 INCLUDED_SOURCE=$(BIRD_SOURCE) $(FELIX_GPL_SOURCE)
 
 # Versions and locations of dependencies used in tests.
-CALICOCTL_VER?=master
-CNI_VER?=master
+CALICOCTL_VER?=v3.14.1
```
caseydavenport, line 57: These should be `master` still, since this is the Makefile for the master branch.

Makefile
```diff
@@ -18,7 +18,7 @@ Makefile.common: Makefile.common.$(MAKE_BRANCH)
 Makefile.common.$(MAKE_BRANCH):
 	# Clean up any files downloaded from other branches so they don't accumulate.
 	rm -f Makefile.common.*
-	curl --fail $(MAKE_REPO)/Makefile.common -o "$@"
+	curl -k --fail $(MAKE_REPO)/Makefile.common -o "$@"
```
caseydavenport, line 21: What is this change for?

Makefile
```diff
@@ -1,5 +1,5 @@
 PACKAGE_NAME?=github.com/projectcalico/node
-GO_BUILD_VER?=v0.45
+GO_BUILD_VER?=v0.40
```
caseydavenport, line 2: I don't think we should revert the makefile version. Why did you need to do this?

Makefile
```diff
@@ -164,6 +165,8 @@ remote-deps: mod-download
 		cp -r `go list -mod=mod -m -f "{{.Dir}}" github.com/projectcalico/felix`/bpf-apache bin/bpf; \
 		chmod -R +w bin/bpf; \
 		chmod +x bin/bpf/bpf-gpl/list-* bin/bpf/bpf-gpl/calculate-*; \
+		sed -i "s/amd64/arm64/" ./bin/bpf/bpf-gpl/Makefile; \
```
caseydavenport, line 168: Hm, I'm not sure if this is a reliable way to go about this.

I think we'd want to enhance that Makefile to have support for multi-arch in a separate PR. It lives here: https://github.com/projectcalico/felix/blob/master/bpf-gpl/Makefile

caseydavenport (changes requested):
Added a few comments. There are some changes that I think aren't related to this PR, some commented lines of code, etc., that I don't fully understand.

Makefile
LyleLee: Agree, the specific v3.14.1 thing is just the tag my special case needs when the latest stable is v3.14.1. A universal match to `master` should be our goal.

Makefile
LyleLee: My internet needs the proxy to go out through, and  `-k` helps me get rip of errors pop out every time I run make.  That was my bad not removing it in the patch.

Makefile
LyleLee: Well, this is just my cook to make it work. Thanks to point out.

Makefile
LyleLee: Totally agree.

Dockerfile.arm64
```diff
@@ -11,39 +11,138 @@
...
 
 # Enable non-native builds of this image on an amd64 hosts.
 # This must be the first RUN command in this file!
 # we only need this for the intermediate "base" image, so we can run all the apk and other commands
 # when running on a kernel >= 4.8, this will become less relevant
-COPY --from=qemu /usr/bin/qemu-aarch64-static /usr/bin/
-
-# Install remaining runtime deps required for felix from the global repository
-RUN apk add --no-cache ip6tables ipset iputils iproute2 conntrack-tools runit ca-certificates
+#COPY --from=qemu /usr/bin/qemu-aarch64-static /usr/bin/
```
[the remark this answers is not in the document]
LyleLee: No, it is not my intent.  it is something left behind because of my rash.

LyleLee:
> Added a few comments. There are some changes that I think aren't related to this PR, some commented lines of code, etc., that I don't fully understand.

I will do my best to submit a new PR, though I do doubt about my ability in this area.  That will be nice if the developer could make this change happen.

caseydavenport:
This PR is a bit out of date now, so I'm going to close it due to inactivity. 

Would be happy to take another look at this if it gets updated.

lwr20:
Likely covered by #1044

PR #549, opened 2020-08-19
WIP: node start-up in run-once and daemon mode

realgaurav:
## Description
node start-up in run-once and daemon mode.

This PR is a WIP with `make default` and `make image` passing atm.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TODO
```

### DIFF
--- a/cmd/calico-node/main.go
+++ b/cmd/calico-node/main.go
@@ -40,7 +40,8 @@ var flagSet = flag.NewFlagSet("Calico", flag.ContinueOnError)
 // Build the set of supported flags.
 var version = flagSet.Bool("v", false, "Display version")
 var runFelix = flagSet.Bool("felix", false, "Run Felix")
-var runStartup = flagSet.Bool("startup", false, "Initialize a new node")
+var runStartup = flagSet.Bool("startup", false, "Initialize a new node with constant monitoring of IP address")
+var startupRunOnce = flagSet.Bool("startup-run-once", false, "Initialize a new node once")
 var runAllocateTunnelAddrs = flagSet.Bool("allocate-tunnel-addrs", false, "Configure tunnel addresses for this node")
 var allocateTunnelAddrsRunOnce = flagSet.Bool("allocate-tunnel-addrs-run-once", false, "Run allocate-tunnel-addrs in oneshot mode")
 var monitorToken = flagSet.Bool("monitor-token", false, "Watch for Kubernetes token changes, update CNI config")
@@ -108,7 +109,7 @@ func main() {
 		felix.Run("/etc/calico/felix.cfg", buildinfo.GitVersion, buildinfo.GitRevision, buildinfo.BuildDate)
 	} else if *runStartup {
 		logrus.SetFormatter(&logutils.Formatter{Component: "startup"})
-		startup.Run()
+		startup.Run(*startupRunOnce)
 	} else if *runConfd {
 		logrus.SetFormatter(&logutils.Formatter{Component: "confd"})
 		cfg, err := confdConfig.InitConfig(true)
--- a/pkg/startup/startup.go
+++ b/pkg/startup/startup.go
@@ -67,6 +67,7 @@ const (
 	KubeadmConfigConfigMap = "kubeadm-config"
 	// Rancher clusters store their state in this config map in the kube-system namespace.
 	RancherStateConfigMap = "full-cluster-state"
+	DEFAULT_AUTODETECT_POLL_INTERVAL = 60 * time.Minute
 )
 
 // Version string, set during build.
@@ -88,7 +89,7 @@ var (
 // -  Configuring the node resource with IP/AS information provided in the
 //    environment, or autodetected.
 // -  Creating default IP Pools for quick-start use
-func Run() {
+func Run(runOnce bool) {
 	// Check $CALICO_STARTUP_LOGLEVEL to capture early log statements
 	configureLogging()
 
@@ -149,7 +150,6 @@ func Run() {
 			} else {
 				log.WithError(err).Error("failed to query kubeadm's config map")
 				terminate()
-				return
 			}
 		}
 
@@ -164,39 +164,15 @@ func Run() {
 			} else {
 				log.WithError(err).Error("failed to query Rancher's cluster state config map")
 				terminate()
-				return
 			}
 		}
 	}
 
-	// Configure and verify the node IP addresses and subnets.
-	checkConflicts, err := configureIPsAndSubnets(node)
-	if err != nil {
-		clearv4 := os.Getenv("IP") == "autodetect"
-		clearv6 := os.Getenv("IP6") == "autodetect"
-		if node.ResourceVersion != "" {
-			// If we're auto-detecting an IP on an existing node and hit an error, clear the previous
-			// IP addresses from the node since they are no longer valid.
-			clearNodeIPs(ctx, cli, node, clearv4, clearv6)
-		}
-		terminate()
-	}
+	checkConflicts := setupIPAndSubnets(ctx, node, cli)
 
 	// If we report an IP change (v4 or v6) we should verify there are no
 	// conflicts between Nodes.
-	if checkConflicts && os.Getenv("DISABLE_NODE_IP_CHECK") != "true" {
-		v4conflict, v6conflict, err := checkConflictingNodes(ctx, cli, node)
-		if err != nil {
-			// If we've auto-detected a new IP address for an existing node that now conflicts, clear the old IP address(es)
-			// from the node in the datastore. This frees the address in case it needs to be used for another node.
-			clearv4 := (os.Getenv("IP") == "autodetect") && v4conflict
-			clearv6 := (os.Getenv("IP6") == "autodetect") && v6conflict
-			if node.ResourceVersion != "" {
-				clearNodeIPs(ctx, cli, node, clearv4, clearv6)
-			}
-
-			terminate()
-		}
+	if checkConflicts {
 	}
 
 	// If Calico is running in policy only mode we don't need to write BGP related details to the Node.
@@ -206,7 +182,7 @@ func Run() {
 
 		if clientset != nil {
 			log.Info("Setting NetworkUnavailable to False")
-			err = setNodeNetworkUnavailableFalse(*clientset, nodeName)
+			err := setNodeNetworkUnavailableFalse(*clientset, nodeName)
 			if err != nil {
 				log.WithError(err).Error("Unable to set NetworkUnavailable to False")
 			}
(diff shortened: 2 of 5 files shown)


pkg/startup/startup.go
```diff
@@ -244,6 +222,61 @@ func Run() {
...
+			autoDetectPollingInterval, _ = time.ParseDuration(os.Getenv("AUTODETECT_POLL_INTERVAL"))
+		}
+		for {
+			select {
+			case <-time.After(autoDetectPollingInterval):
+				checkConflicts = setupIPAndSubnets(ctx, node, cli)
+				if checkConflicts {
+					checkNodeIPConflictCheck(ctx, node, cli)
+					// Apply the updated node resource.
+					if _, err := CreateOrUpdate(ctx, cli, node); err != nil {
```
caseydavenport, line 238: We need to do this conditionally - namely, we don't want to update the node object if it hasn't changed, so we should compare to see if we actually need to make the API request.

pkg/startup/startup.go
```diff
@@ -244,6 +222,61 @@ func Run() {
 		log.WithError(err).Errorf("Unable to ensure network for os")
 		terminate()
 	}
+
+	if !runOnce {
```
caseydavenport, line 226: Looking at this, I'm starting to think that this belongs separately from `startup.go`. 

First thing I'm thinking is that it doesn't make sense to have this code run multiple times:
- https://github.com/projectcalico/node/blob/87af881af17d24f785309cc43883b59778d51c85/pkg/startup/startup.go#L94-L118

And also this code should only need to run at startup, and will never need to be periodic:
- https://github.com/projectcalico/node/blob/87af881af17d24f785309cc43883b59778d51c85/pkg/startup/startup.go#L124-L170

Configuring the AS number will never change, since it's done based on an env var so needs a pod restart anyway:
- https://github.com/projectcalico/node/blob/87af881af17d24f785309cc43883b59778d51c85/pkg/startup/startup.go#L183

And same for configuring IP pools:
- https://github.com/projectcalico/node/blob/87af881af17d24f785309cc43883b59778d51c85/pkg/startup/startup.go#L207-L224

Now, I understand that this PR doesn't include those in the periodic / poll loop. But, it does introduce a behavior where all of that now gets executed _twice_ per node launch (once in the initial call to startup, and once when the startup process is launched).

I think we should have a new routine for the periodic code so that it doesn't call all of that again (and we can name it something other than startup, since it's not just at startup anymore). Most of what you've done here would be the same, but we'd want to commonize/export the IP auto-detection function so it could be used in both places. WDYT?

pkg/startup/startup.go
```diff
@@ -164,39 +165,16 @@ func Run() {
 			} else {
 				log.WithError(err).Error("failed to query Rancher's cluster state config map")
 				terminate()
-				return
 			}
 		}
 	}
 
-	// Configure and verify the node IP addresses and subnets.
```
caseydavenport, line 172: If we're putting these into helper functions, any reason it needs to be two separate functions rather than just a single function?

pkg/startup/startup.go
```diff
@@ -244,6 +222,61 @@ func Run() {
...
+	checkConflicts, err := configureIPsAndSubnets(node)
+	if err != nil {
+		clearv4 := os.Getenv("IP") == "autodetect"
+		clearv6 := os.Getenv("IP6") == "autodetect"
+		if node.ResourceVersion != "" {
+			// If we're auto-detecting an IP on an existing node and hit an error, clear the previous
+			// IP addresses from the node since they are no longer valid.
+			clearNodeIPs(ctx, cli, node, clearv4, clearv6)
+		}
+		terminate()
```
caseydavenport, line 259: We need to be careful about the use of `terminate()` here - I don't think we want to terminate in the daemon version of this code. Rather we should return an error and catch it in a retry loop of some kind.

I guess if we terminate then `runsv` will start the service again from the beginning, so maybe not the end of the world.

pkg/startup/startup.go
```diff
@@ -66,7 +67,8 @@ const (
 	// KubeadmConfigConfigMap is defined in k8s.io/kubernetes, which we can't import due to versioning issues.
 	KubeadmConfigConfigMap = "kubeadm-config"
 	// Rancher clusters store their state in this config map in the kube-system namespace.
-	RancherStateConfigMap = "full-cluster-state"
+	RancherStateConfigMap            = "full-cluster-state"
+	DEFAULT_AUTODETECT_POLL_INTERVAL = 60 * time.Minute
```
caseydavenport, line 71: I think this interval is probably too long. This means that potentially, when a node gets a new address over DHCP, that node would be out of operation for an hour. 

I was imagining a polling interval more on the order of 10 seconds or so so that changes are detected quickly with minimal interruption. There's not really a way to get _no_ interruption, because the address for the node is changing and routes will be need to be updated, but 10 seconds or so seems acceptable.

caseydavenport: Ideally we would be monitoring using something like this and only reconciling when we get a notification that an address has changed on the host: https://godoc.org/github.com/vishvananda/netlink#AddrSubscribe

realgaurav: Thanks Casey. The latest change addresses this comment. PTAL.

pkg/startup/startup.go
realgaurav: added a separate function to address this comment.

pkg/startup/startup_windows.go
```diff
@@ -96,3 +96,7 @@ func ensureNetworkForOS(ctx context.Context, c client.Interface, nodeName string
 	logrus.Info("Ensure network is done.")
 	return nil
 }
+
+// Placeholder to compile windows binary.
+func MonitorIPAddressSubnets() {
```
caseydavenport, line 101: I think this implementation is wrong for Windows - rather than exit immediately, we need this function to "sleep" forever.

You could have it trying to a read from a channel with no producers, for example.

pkg/startup/startup_linux.go
```diff
@@ -100,3 +103,41 @@ func ipv6Supported() bool {
...
+	}
+
+	ctx := context.Background()
+	_, cli := calicoclient.CreateClient()
+	nodeName := determineNodeName()
+	node := getNode(ctx, cli, nodeName)
+
+	for {
+		select {
+		case <-time.After(autoDetectPollingInterval):
```
caseydavenport, line 127: case statements don't automatically fall through in golang, so you'll need an explicit `fallthrough` keyword here to make it do so.

Also probably worth a log stating that this resync has been triggered by a periodic timer, rather than an interface update.

pkg/startup/startup_linux.go
```diff
@@ -100,3 +103,41 @@ func ipv6Supported() bool {
...
+func MonitorIPAddressSubnets() {
+	autoDetectPollingInterval := DEFAULT_AUTODETECT_POLL_INTERVAL
+	if os.Getenv("AUTODETECT_POLL_INTERVAL") != "" {
+		autoDetectPollingInterval, _ = time.ParseDuration(os.Getenv("AUTODETECT_POLL_INTERVAL"))
+	}
+
+	// Add a subscription to get updated if there is any change to interface addresses.
+	addrUpdate := make(chan netlink.AddrUpdate)
+	done := make(chan struct{})
+	if err := netlink.AddrSubscribe(addrUpdate, done); err != nil {
```
caseydavenport, line 116: Did you confirm if this monitors for just IP addresses in the root network namespace, or will we get addr notifications for other netns as well?

It would be bad, for example, if we got notified if pod addresses coming and going :D

pkg/startup/startup_linux.go
```diff
@@ -100,3 +103,41 @@ func ipv6Supported() bool {
...
+		log.WithError(err).Error("Failed to subscribe to network interface updates")
```
caseydavenport, line 117: If we hit this error do we want to exit?

pkg/startup/startup_windows.go
realgaurav: I was thinking, for lake of monitoring IP address/subnet in Windows, to retain the old behavior which is similar to `run-once`. I have 2 questions --

1. what's the advantage of 'sleep' for Windows when we are not monitoring the IP?
2. should we make this -- autodetect based on time alone for Windows?

pkg/startup/startup.go
```diff
@@ -246,6 +245,23 @@ func Run() {
 	}
 }
 
+func checkIPAddressSubnets(ctx context.Context, node *api.Node, cli client.Interface) bool {
```
caseydavenport, line 248: I think this function needs to respect the user-provided auto-detection logic. 

e.g., this line here: https://github.com/projectcalico/node/blob/99c269555fdab674f32a5087c310b1804a622033/pkg/startup/startup.go#L467

Is there a reason we can't just call this function: https://github.com/projectcalico/node/blob/99c269555fdab674f32a5087c310b1804a622033/pkg/startup/startup.go#L448?

Also looks like `ctx` and `cli` are unused, so don't need to be here.

pkg/startup/startup_windows.go
caseydavenport: We could do the second option, yeah.

The main thing is that if we exit, then the service will have "crashed" and runsv will attempt to restart the service, which of course will then exit immediately.

realgaurav: I will add auto-detect based on time for Windows.

pkg/startup/startup_linux.go
realgaurav: I wasn't sure if exit() would resolve issue with AddrSubscribe() failure, so proposed to log the error and proceed with time-out logic. If the error is unrecoverable, startup will get stuck into a restart loop. Let me know what you think.

pkg/startup/startup.go
realgaurav: My original proposal called `configureIPsAndSubnets` directly. However, I noticed the following condition which looked problematic from monitoring perspective --

`} else if ipv4Env == "none" && node.Spec.BGP.IPv4Address != "" {`

If the IP address is already set which is try for monitoring case, having `ipv4Env == "none"` will just do `validateIP` and `terminate` if the IP address is not available. I have seen `ipv4Env = "none"` in a few cases (e.g. AKS) which will lead to a restart.

Should we just ignore monitoring if `ipv4Env` is set to `none`?

pkg/startup/startup_linux.go
caseydavenport: Yeah, good point. I'm not sure what the "best" behavior is here.

cmd/calico-node/main.go
```diff
@@ -40,7 +40,8 @@ var flagSet = flag.NewFlagSet("Calico", flag.ContinueOnError)
 // Build the set of supported flags.
 var version = flagSet.Bool("v", false, "Display version")
 var runFelix = flagSet.Bool("felix", false, "Run Felix")
-var runStartup = flagSet.Bool("startup", false, "Initialize a new node")
+var runStartup = flagSet.Bool("startup", false, "Initialize a new node with constant monitoring of IP address")
```
caseydavenport, line 43: Sorta a nit, but I think it's fair to say that these are really separate options.

I'd recommend

```
flagSet.Bool("startup", false, "Initialize a new node")
flagSet.Bool("monitor-addresses", "false", "Monitor for node IP address changes")
```

realgaurav: okie.

pkg/startup/startup_linux.go
realgaurav: It's bad :grin: 

I played around a bit with a stub program but didn't verify behavior with transient container. With `AddrSubscribe`, we get notified as soon as an container is created or destroyed. We can do better in terms of identifying new addresses only but IIU your comment correctly, it's equally bad.

I will propose the following improvement to make it acceptable --
We start with a list of available interfaces and build a map of link-indexs (with link-name which is not usable atm) and once we receive an address update we check if the update corresponds to any of the available interface. We log a new interface update but do nothing further. Only when the address update corresponds to the available interfaces, we will check for the node address change. I sort-of verified that the address update coming for a transient container doesn't match the list we started with using a stub program. Let me know what you think.

Thanks

pkg/startup/startup.go
caseydavenport: Ah, hm. Yeah, I hadn't spotted that `validateIP` eventually can call terminate.

Ideally we keep the logic as close to the same as possible - I want to avoid accidental scenarios where the startup script does one thing and then this continuous code does something subtly different.

Maybe we need to do a minor refactor so that `validateIP` doesn't immediately terminate? Or we can pass a "terminationAllowed" bool through? 

Or maybe we should just pull `validateIP` out of that function altogether, and call `validateIP` after the `configureIPsAndSubnets` call exits?

We would also need to deal with this line somehow: https://github.com/projectcalico/node/blob/99c269555fdab674f32a5087c310b1804a622033/pkg/startup/startup.go#L524

caseydavenport: Thinking about this a bit more, we probably _also_ need to respect the collision detection logic here: https://github.com/projectcalico/node/blob/99c269555fdab674f32a5087c310b1804a622033/pkg/startup/startup.go#L187-L188

If our IP address changes, and we notice that it is the same as another node, that is a problem, and we shouldn't write the new addresses to the data store just yet. However, we should also clear our the old addresses, because we've determined that they aren't valid any more, so keeping them set would (1) be wrong and (2) potentially prevent other nodes from rightfully claiming them (e.g., if another node is assigned the address that this node used to have).

pkg/startup/startup_linux.go
caseydavenport: Did you try using `AddrSubscribeAt` and passing in the current netns via `netns.Get()`? 

https://godoc.org/github.com/vishvananda/netns#Get

PR #564, opened 2020-09-01
AUTODETECTION_METHOD_K8S option

tomastigera:
When running in kubernetes environment, provide the option to detect the
node addresses from the k8s node resource. This may give the most precise
autodetection option when in an environment with multiple NICs. This is
independent of the interface naming or reachability and is well suited
for runnig in policy only mode, e.g. on EKS with AWS CNI.

Precise node IP detection is necessary for eBPF to program its routes,
which then used, for instance, for connect time resolution of nodeports
both by the pods and hosts.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/startup/startup.go
+++ b/pkg/startup/startup.go
@@ -63,6 +63,7 @@ const (
 	AUTODETECTION_METHOD_INTERFACE      = "interface="
 	AUTODETECTION_METHOD_SKIP_INTERFACE = "skip-interface="
 	AUTODETECTION_METHOD_CIDR           = "cidr="
+	AUTODETECTION_METHOD_K8S            = "k8s"
 
 	DEFAULT_MONITOR_IP_POLL_INTERVAL = 60 * time.Second
 
@@ -170,7 +171,7 @@ func Run() {
 		}
 	}
 
-	configureAndCheckIPAddressSubnets(ctx, cli, node)
+	configureAndCheckIPAddressSubnets(ctx, cli, node, clientset)
 
 	// If Calico is running in policy only mode we don't need to write BGP related details to the Node.
 	if os.Getenv("CALICO_NETWORKING_BACKEND") != "none" {
@@ -234,9 +235,14 @@ func getMonitorPollInterval() time.Duration {
 	return interval
 }
 
-func configureAndCheckIPAddressSubnets(ctx context.Context, cli client.Interface, node *api.Node) bool {
+// configureAndCheckIPAddressSubnets uses k8sClient iff autodetection uses k8s
+// but is plombed through here so that an existing client can be reused. It can
+// be nil.
+func configureAndCheckIPAddressSubnets(ctx context.Context, cli client.Interface,
+	node *api.Node, k8sClient kubernetes.Interface) bool {
+
 	// Configure and verify the node IP addresses and subnets.
-	checkConflicts, err := configureIPsAndSubnets(node)
+	checkConflicts, err := configureIPsAndSubnets(node, k8sClient)
 	if err != nil {
 		// If this is auto-detection error, do a cleanup before returning
 		clearv4 := os.Getenv("IP") == "autodetect"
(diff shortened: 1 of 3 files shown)


pkg/startup/startup.go
```diff
@@ -63,6 +63,7 @@ const (
 	AUTODETECTION_METHOD_INTERFACE      = "interface="
 	AUTODETECTION_METHOD_SKIP_INTERFACE = "skip-interface="
 	AUTODETECTION_METHOD_CIDR           = "cidr="
+	AUTODETECTION_METHOD_K8S            = "k8s"
```
fasaxc, line 66: For future-proofing, should it be `k8s=internal-ip` or something like that?

tomastigera: In the next step, we are also going to get the external IP. We pretty much require that knowledge in EKS. But we canmake it configurable this way. `k8s=internal` and `k8s=all` in the next step. WDYT?

pkg/startup/startup.go
```diff
@@ -687,11 +715,35 @@ func evaluateENVBool(envVar string, defaultValue bool) bool {
...
+	log.Infof("IP_AUTODETECTION_METHOD=%s", method)
+
 	if method == "" || method == AUTODETECTION_METHOD_FIRST {
 		// Autodetect the IP by enumerating all interfaces (excluding
 		// known internal interfaces).
 		return autoDetectCIDRFirstFound(version)
+	} else if method == AUTODETECTION_METHOD_K8S {
+		if k8sClient == nil {
+			log.Errorf("Invalid k8s client for IP autodetection method: %s", method)
+			return nil
```
caseydavenport, line 729: I think in most other places we call `terminate()` for errors like this?

pkg/startup/startup.go
```diff
@@ -687,11 +715,35 @@ func evaluateENVBool(envVar string, defaultValue bool) bool {
 
 // autoDetectCIDR auto-detects the IP and Network using the requested
 // detection method.
-func autoDetectCIDR(method string, version int) *cnet.IPNet {
+func autoDetectCIDR(method string, version int,
+	k8sClient kubernetes.Interface, nodeName string) *cnet.IPNet {
```
caseydavenport, line 719: IMO this can be a single line. We don't have line limitations in Calico and its easier to read on one line than on two.

pkg/startup/startup.go
```diff
@@ -687,11 +715,35 @@ func evaluateENVBool(envVar string, defaultValue bool) bool {
...
+			return nil
+		}
+
+		ips := autodetection.K8sNodeInternalIPs(k8sNode, version)
+		if len(ips) == 0 {
+			log.Errorf("No k8s internal IP version %d", version)
+			return nil
+		}
+		if len(ips) > 1 {
+			log.Warnf("More then 1 k8s internal IP version %d, picking one: %s", version, ips[0])
```
caseydavenport, line 744: Do we have any future plans to improve on this? Do we need / want a way to limit the address selection even further?

pkg/startup/startup.go
```diff
@@ -234,9 +235,14 @@ func getMonitorPollInterval() time.Duration {
 	return interval
 }
 
-func configureAndCheckIPAddressSubnets(ctx context.Context, cli client.Interface, node *api.Node) bool {
+// configureAndCheckIPAddressSubnets uses k8sClient iff autodetection uses k8s
+// but is plombed through here so that an existing client can be reused. It can
```
caseydavenport, line 239: plombed -> plumbed.

pkg/startup/startup.go
tomastigera: The future plan we pondered with @fasaxc is to introduce a field (in the next step to add external IP) that could hold any number of addresses and keep the current single-address field for compatibility. That would also address the issue with possible multiple internal IPs in k8s.

This PR would make IP detection in k8s deterministic with minimal change and the rest being a gradual improvement on it.

pkg/startup/startup.go
caseydavenport: Agree we should start with `k8s=internal` and then add in additional methods.

If we start with just `k8s` then we'll need to maintain support for that field indefinitely. We should try to agree all of the options that we intend to write up-front.

pkg/startup/startup.go
caseydavenport: That is probably a good thing to do. It is an API change though so we need to be careful about spelling out exactly how the field looks and behaves in various scenarios, how upgrade works, etc.  

Personally, I'd prefer if we have that design agreed before we make intermediate steps towards it so that we don't make any missteps.

tomastigera:
After a discussion with @caseydavenport  and @fasaxc we decided to populate (in KDD mode) a list of addresses from the existing fields and from other fields in the k8s node, therefore a new autodetection in k8s does not seem to be needed at this moment.

PR #567, opened 2020-09-08
Auto detect Calico path for Windows.

song-jiang:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/windows-packaging/CalicoWindows/kubernetes/kube-proxy-service.ps1
+++ b/windows-packaging/CalicoWindows/kubernetes/kube-proxy-service.ps1
@@ -16,10 +16,12 @@ Param(
     [string]$NetworkName = "Calico"
 )
 
-# Import HNS libraries, included in the package.
-ipmo -Force c:\CalicoWindows\libs\hns\hns.psm1
+$baseDir = "$PSScriptRoot\.."
 
-. c:\CalicoWindows\config.ps1
+# Import HNS libraries, included in the package.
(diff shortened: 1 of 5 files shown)


windows-packaging/CalicoWindows/kubernetes/kube-proxy-service.ps1
```diff
@@ -16,10 +16,12 @@ Param(
     [string]$NetworkName = "Calico"
 )
 
+$CalicoPath=(Resolve-Path ..).Path
```
lmm, line 19: I wonder if we need to ensure that this script will be called from the directory we expect.

Maybe we need to check if https://github.com/projectcalico/node/blob/master/windows-packaging/CalicoWindows/kubernetes/install-kube-services.ps1 is run from the right directory (`c:\CalicoWindows/kubernetes/`)

lmm: IIUC Resolve-Path gets the value `..` relative to the current directory that the script is invoked from

song-jiang: We are using relative directory here because the assumption is `kubernetes` directory located at the same level with `config.ps1` and other Calico files. All the files been unzipped from a single archive so it should always be running from the right directory.

lmm (approved):

PR #569, opened 2020-09-09
Conditionally terminate if IP isn't available under calico/node resource.

realgaurav:
## Description

Conditionally terminate if IP isn't available under calico/node resource.
If calico isn't being used for networking, don't restart when IP address is not available under the resource.

## Todos
- [x] Tests: UT + tested in an EKS cluster (aws-vpc CNI) that calico-node doesn't crash.
- [ ] Documentation

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/startup/startup.go
+++ b/pkg/startup/startup.go
@@ -256,7 +256,7 @@ func configureAndCheckIPAddressSubnets(ctx context.Context, cli client.Interface
 			// Unrecoverable error, terminate to restart.
 			terminate()
 		} else {
-			log.Warn("No IPv4 or IPv6 addresses configured, and autodetection is not enabled. Some features may not work properly.")
+			log.Warn("No IPv4 or IPv6 addresses configured or detected. Some features may not work properly.")
 			// Bail here setting BGPSpec to nil (if empty) to pass validation.
 			if reflect.DeepEqual(node.Spec.BGP, &api.NodeBGPSpec{}) {
 				node.Spec.BGP = nil
--- a/pkg/startup/startup.go
+++ b/pkg/startup/startup.go
@@ -250,19 +250,6 @@ func configureAndCheckIPAddressSubnets(ctx context.Context, cli client.Interface
 		terminate()
 	}
 
-	if node.Spec.BGP.IPv4Address == "" && node.Spec.BGP.IPv6Address == "" {
-		if os.Getenv("CALICO_NETWORKING_BACKEND") != "none" {
-			log.Error("No IP Addresses configured, and autodetection is not enabled")
-			// Unrecoverable if Calico is handling networking, terminate to restart.
-			terminate()
-		} else {
-			log.Warn("No IP Addresses configured, and autodetection is not enabled. Some features may not work properly.")
-			// Bail here setting node.Spec.BGP to nil so that the node validation passes.
-			node.Spec.BGP = nil
-			return checkConflicts
-		}
-	}
-
 	// If we report an IP change (v4 or v6) we should verify there are no
 	// conflicts between Nodes.
 	if checkConflicts && os.Getenv("DISABLE_NODE_IP_CHECK") != "true" {
@@ -564,6 +551,11 @@ func configureIPsAndSubnets(node *api.Node) (bool, error) {
 			node.Spec.BGP.IPv6Address = parseIPEnvironment("IP6", ipv6Env, 6)
 		}
 		validateIP(node.Spec.BGP.IPv6Address)
+	}
+
+	if ipv4Env == "none" && (ipv6Env == "" || ipv6Env == "none") && node.Spec.BGP.IPv4Address == "" && node.Spec.BGP.IPv6Address == "" {
+		log.Warn("No IP Addresses configured, and autodetection is not enabled")
+		terminate()
 	}
 
 	// Detect if we've seen the IP address change, and flag that we need to check for conflicting Nodes
--- a/pkg/startup/startup.go
+++ b/pkg/startup/startup.go
@@ -250,6 +250,19 @@ func configureAndCheckIPAddressSubnets(ctx context.Context, cli client.Interface
 		terminate()
 	}
 
+	if node.Spec.BGP.IPv4Address == "" && node.Spec.BGP.IPv6Address == "" {
+		if os.Getenv("CALICO_NETWORKING_BACKEND") != "none" {
+			log.Error("No IP Addresses configured, and autodetection is not enabled")
+			// Unrecoverable if Calico is handling networking, terminate to restart.
+			terminate()
+		} else {
+			log.Warn("No IP Addresses configured, and autodetection is not enabled. Some features may not work properly.")
+			// Bail here setting node.Spec.BGP to nil so that the node validation passes.
+			node.Spec.BGP = nil
+			return checkConflicts
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


pkg/startup/startup.go
```diff
@@ -555,7 +555,11 @@ func configureIPsAndSubnets(node *api.Node) (bool, error) {
 
 	if ipv4Env == "none" && (ipv6Env == "" || ipv6Env == "none") && node.Spec.BGP.IPv4Address == "" && node.Spec.BGP.IPv6Address == "" {
 		log.Warn("No IP Addresses configured, and autodetection is not enabled")
```
caseydavenport, line 557: I think we should have two variants of this warning log.

One within the CALICO_NETWORKING_BACKEND branch, which is an error level log

One outside of it, that is Warning level and says something like this:

> No IP Addresses configured, and autodetection is not enabled. Some features may not work properly.

pkg/startup/startup.go
```diff
@@ -555,7 +555,11 @@ func configureIPsAndSubnets(node *api.Node) (bool, error) {
...
-		terminate()
+		// Unrecoverable if Calico is handling networking, terminate to restart.
+		if os.Getenv("CALICO_NETWORKING_BACKEND") != "none" {
+			terminate()
+		}
+		return false, nil
```
robbrockbank, line 562: I don't think you should return false here - strictly according to the method doc we should still run through the code below to determine if the IPs have changed.

pkg/startup/startup.go
```diff
@@ -555,7 +555,11 @@ func configureIPsAndSubnets(node *api.Node) (bool, error) {
 
 	if ipv4Env == "none" && (ipv6Env == "" || ipv6Env == "none") && node.Spec.BGP.IPv4Address == "" && node.Spec.BGP.IPv6Address == "" {
 		log.Warn("No IP Addresses configured, and autodetection is not enabled")
-		terminate()
+		// Unrecoverable if Calico is handling networking, terminate to restart.
```
robbrockbank, line 558: Not fully sure I follow the check on line 556.

Maybe we could move this block out into the calling function as part of the post-configureIPs check.  In the calling function I think you'd just need to check that NETWORKING BACKEND is none and both IPs are blank.  If so terminate.

I don't think it's necessary to check the ipv4Env and ipv6Env values are none or blank.  If they are not set to none (or blank) then we'll hit an error if we don't manage to extract an IP from the value, and we only really care if we don't have any IPs and we are using calico networking.

pkg/startup/startup.go
realgaurav: hmm, but that means if an IP address is empty (which is OK now in some cases) will be compared to previously set value. Is that necessary? What condition are you thinking this will be necessary to IP address comparison for?

pkg/startup/startup.go
realgaurav: I would read :556 as neither IPv4 or IPv6 addresses are available under calico/node resource.

Do you mean? do the following after calling `configureIPsAndSubnets`:

```
if node.Spec.BGP.IPv4Address == "" && node.Spec.BGP.IPv6Address == "" {
    if os.Getenv("CALICO_NETWORKING_BACKEND") != "none" {
        // log-1
	terminate()
     } else {
        // log-2
     } 
}
```

if yes, question in my earlier comment is relevant to this proposal.

Thanks

pkg/startup/startup.go
robbrockbank: I don't really have anything in mind - other than trying to simplify the flow and keeping the behavior true to the function signature.

pkg/startup/startup.go
robbrockbank: Yup that's what I was thinking.  It feels easier to following having the "assign IPs" separate from the "check IP config is consistent with other config".

pkg/startup/startup.go
```diff
@@ -250,6 +250,19 @@ func configureAndCheckIPAddressSubnets(ctx context.Context, cli client.Interface
 		terminate()
 	}
 
+	if node.Spec.BGP.IPv4Address == "" && node.Spec.BGP.IPv6Address == "" {
+		if os.Getenv("CALICO_NETWORKING_BACKEND") != "none" {
+			log.Error("No IP Addresses configured, and autodetection is not enabled")
```
caseydavenport, line 255: I don't think this error makes sense any more - we moved the check and it no longer knows if auto-detection is enabled or not. All it knows is that there are no addresses and that it needs them.

I think we need to reword the error if we're going to move it out here.

e.g., "No IP addresses configured or detected, but they are required for Calico networking"

caseydavenport: Still thinking through whether it's OK to actually move this check here.

realgaurav: Sure, I have adjusted the UT with this change. Let me know if it makes sense given the usage there.

pkg/startup/startup.go
```diff
@@ -250,6 +250,19 @@ func configureAndCheckIPAddressSubnets(ctx context.Context, cli client.Interface
...
+			// Unrecoverable if Calico is handling networking, terminate to restart.
+			terminate()
+		} else {
+			log.Warn("No IP Addresses configured, and autodetection is not enabled. Some features may not work properly.")
+			// Bail here setting node.Spec.BGP to nil so that the node validation passes.
+			node.Spec.BGP = nil
```
caseydavenport, line 261: I don't think this is the right action. The node BGP spec has a bunch of other fields in it that we don't just want to overwrite:

https://github.com/projectcalico/libcalico-go/blob/master/lib/apis/v3/node.go#L78-L93

I think we need to compare the BGP spec to an "empty" one, and if it is empty, then set it to nil.

realgaurav: I wasn't able to think why this field would be populated for non-Calico networking cases. But, if we have to support what you are saying, I can do the empty comparison.

pkg/startup/startup.go
caseydavenport: The fields themselves aren't used if not using Calico networking. However, it's bad form to silently modify the user's intent (e.g., if a user sets one of these fields, and we allow it, we shouldn't come along and clear them out later).

One somewhat contrived scenario might be that they are planning on switching to Calico networking and this is a transient state.

pkg/startup/startup.go
realgaurav: I have added the BPGSpec comparison before settting it to `nil`

pkg/startup/startup.go
```diff
@@ -250,6 +250,21 @@ func configureAndCheckIPAddressSubnets(ctx context.Context, cli client.Interface
 		terminate()
 	}
 
+	if node.Spec.BGP.IPv4Address == "" && node.Spec.BGP.IPv6Address == "" {
+		if os.Getenv("CALICO_NETWORKING_BACKEND") != "none" {
+			log.Error("No IPv4 or IPv6 addresses configured or detected, required for Calico networking")
+			// Unrecoverable error, terminate to restart.
+			terminate()
+		} else {
+			log.Warn("No IPv4 or IPv6 addresses configured, and autodetection is not enabled. Some features may not work properly.")
```
caseydavenport, line 259: this log still refers to "autodetction", I think it should mirror the one above (configured or detected.)

caseydavenport (approved):
One minor comment but otherwise lgtm

PR #583, opened 2020-10-06
Fix DNS resolution in calico-node containers

NeonSludge:
## Description
The `calico/node` image is missing the `libresolv.so.2` library: it gets deleted by the `clean-up-filesystem.sh` script because it's not directly linked to any binaries from the "keep" list.

This library is provided by the `glibc` package and is necessary for DNS resolution to work.
Without it, every `getaddrinfo()` call returns -11 (`EAI_SYSTEM`) and no names can be resolved from inside the `calico-node` containers.

If Calico is configured to use an external etcdv3 cluster as its datastore, this makes it impossible to use names instead of IP addresses in the etcd endpoint list.

This PR adds `libresolv` to the "well-known plugins" list so that it doesn't get cleaned up.

## Release Note

```release-note
Fix DNS resolution in calico-node containers.
```

### DIFF
--- a/clean-up-filesystem.sh
+++ b/clean-up-filesystem.sh
@@ -213,7 +213,7 @@ while read -r path; do
     continue
   fi
   # Well-known plugins, not directly linked.
-  if [[ "$path" =~ xtables|netfilter|conntrack|ct_|pam|libnss ]] && ! [[ "$path" =~ systemd ]] ; then
+  if [[ "$path" =~ xtables|netfilter|conntrack|ct_|pam|libnss|libresolv ]] && ! [[ "$path" =~ systemd ]] ; then
     echo "PLUGIN: $path"
     libs_to_keep[$path]=true
     continue


fasaxc (approved):

fasaxc:
Great, thanks for figuring this out.  I'm pretty sure that some cases work without this change but maybe it's needed for SRV records or something?

NeonSludge:
> I'm pretty sure that some cases work without this change but maybe it's needed for SRV records or something?

This breaks retrieving A records as well.
We had a simple setup with a k8s cluster, an external etcd cluster, an external DNS cluster and hostnames in the etcd endpoint URL list for calico-node. Those hostnames were supposed to be resolved via that DNS server.

Without this fix, calico-node gets stuck connecting to the datastore because it can't resolve the hostnames. We figured this out by tracing system and library calls made by the calico-node process.
After we switched to a custom calico-node image with the libresolv library present, everything went back to normal.

PR #589, opened 2020-10-12
windows: kube-services scripts support single service

lmm:
Description

This PR makes the kube-services install/uninstall scripts also support a single service. Some platforms only need kube-proxy installed by our scripts. The in-progress OCP+CalicoWindows installation docs PR uses install-kube-proxy.ps1: https://github.com/projectcalico/calico/pull/4058/commits/c3109661867c96ca67a3a0813c61e7b8e41e03b8

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
```

### DIFF
--- a/windows-packaging/CalicoWindows/kubernetes/install-kube-services.ps1
+++ b/windows-packaging/CalicoWindows/kubernetes/install-kube-services.ps1
@@ -12,6 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+param
+(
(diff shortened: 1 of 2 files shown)


song-jiang (changes requested):
I would suggest we keep current file structure and use parameters to run separate services. With no parameter provided ( as default) , both kubelet/kube-proxy will be handled.

e.g. If user want to install kube-proxy only, they will do `install-kube-services -service kube-proxy` .

song-jiang (approved):
LGTM

Issue #563, opened 2020-08-29
(v3.13.4)IPIP set cross-subnet, nodes in the same cidr still uses tunl0 iface to communicate


njuptlzf:
IPIP set cross-subnet, nodes in the same cidr still uses tunl0 iface to communicate

## Expected Behavior

node in the same cdir communicates through the env IP interface(ens8)
node in the different cdir communicates through the tunl0 interface

## Current Behavior

$ route -n
Kernel IP routing table
```
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
0.0.0.0         192.168.122.1   0.0.0.0         UG    0      0        0 ens8
169.254.0.0     0.0.0.0         255.255.0.0     U     1003   0        0 ens8
169.254.0.0     0.0.0.0         255.255.0.0     U     1004   0        0 ens9
172.17.0.0      0.0.0.0         255.255.0.0     U     0      0        0 docker0
172.22.73.0     192.168.122.70  255.255.255.192 UG    0      0        0 tunl0    <------------
172.22.132.192  0.0.0.0         255.255.255.192 U     0      0        0 *
172.22.132.196  0.0.0.0         255.255.255.255 UH    0      0        0 calibdf597a306f
172.22.155.0    192.168.122.71  255.255.255.192 UG    0      0        0 tunl0    <-----------
192.168.122.0   0.0.0.0         255.255.255.0   U     0      0        0 ens8

```
env:
```
cat /etc/kubernetes/calico_env/env

IP=192.168.122.69
IP6=none
```

calico-node config:
```
containers:
  - args:
    - cat /etc/calico/env/env && export $(cat /etc/calico/env/env) && start_runit
    command:
    - /bin/sh
    - -c
    env:
    - name: CALICO_IPV4POOL_CIDR
      value: 172.22.0.0/16
    - name: CALICO_IPV6POOL_CIDR
      value: fd03::0/112
    - name: CALICO_IPV4POOL_IPIP
      value: cross-subnet            <------------
    - name: FELIX_IPV6SUPPORT
      value: "true"
    - name: CALICO_IPV6POOL_NAT_OUTGOING
      value: "true"
    - name: DATASTORE_TYPE
      value: kubernetes
    - name: WAIT_FOR_DATASTORE
      value: "true"
    - name: NODENAME
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: spec.nodeName
    - name: CALICO_NETWORKING_BACKEND
      valueFrom:
[...]

PR #593, opened 2020-10-17
Automatically find the cidr of IPv4Address when IP is ipStr

njuptlzf:
## Description

Fixes #563
When calico is used as a network plugin for kubernets, for convenience, only calico-cni.yml is used to configure the environment variables of calico-node.
```
- name: IP
 
  valueFrom:
 
    fieldRef:
 
      fieldPath: status.hostIP

```

At this time, it is hoped that calico-node can automatically discover the local cidr, so as to avoid using Block address and use IPIP successfully.
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/startup/startup.go
+++ b/pkg/startup/startup.go
@@ -538,7 +538,12 @@ func configureIPsAndSubnets(node *api.Node) (bool, error) {
 		validateIP(node.Spec.BGP.IPv4Address)
 	} else if ipv4Env != "none" {
 		if ipv4Env != "" {
-			node.Spec.BGP.IPv4Address = parseIPEnvironment("IP", ipv4Env, 4)
+			// Attempt to get the local CIDR of ipv4Env
+			ipv4CIDROrIP, err := getLocalCIDR(ipv4Env, 4, autodetection.GetInterfaces)
+			if err != nil {
+				log.Warnf("Attempt to get the local CIDR: %s failed, %s", ipv4Env, err)
+			}
+			node.Spec.BGP.IPv4Address = parseIPEnvironment("IP", ipv4CIDROrIP, 4)
 		}
 		validateIP(node.Spec.BGP.IPv4Address)
 	}
@@ -1375,4 +1380,49 @@ func extractKubeadmCIDRs(kubeadmConfig *v1.ConfigMap) (string, string, error) {
 	}
 
 	return v4, v6, err
+}
+
+// getLocalCIDR attempts to merge CIDR information from the host with the given IP address.
+// If a CIDR is provided, then it is simply returned.
+// If an IP is provided, it attempts to find the matching interface on the host to detect the appropriate prefix length.
+// If no match is found, the IP will be returned unmodified.
+func getLocalCIDR(ip string, version int, getInterfaces func([]string, []string, int) ([]autodetection.Interface, error)) (string, error) {
+	if strings.Contains(ip, "/") {
+		// Already a CIDR
+		return ip, nil
+	}
+
+	var destCIDR net.IP
+	if version == 4 {
+		destCIDR = net.ParseIP(ip).To4()
+	} else {
+		destCIDR = net.ParseIP(ip).To16()
+	}
+
+	if destCIDR == nil {
+		return ip, fmt.Errorf("%s is invalid.", ip)
+	}
+
+	// Get a full list of interface and IPs and find the CIDR matching the
+	// found IP.
+	interfaces, err := getInterfaces(nil, nil, version)
+	if err != nil {
+		return ip, err
+	}
+
+	log.Debugf("Auto-detecting IPv%d CIDR of %s", version, ip)
+	for _, iface := range interfaces {
+		log.WithField("Name", iface.Name).Debug("Checking interface")
+		for _, cidr := range iface.Cidrs {
+			log.WithField("CIDR", cidr.String()).Debug("Found")
+			if cidr.IP.Equal(destCIDR) {
+				log.WithField("CIDR", cidr.String()).Info("Including CIDR information from host interface.")
+				return cidr.String(), nil
+			}
+		}
+	}
+
+	// Even if no CIDR is found, it doesn't think it needs to throw an exception
+	log.Warnf("Unable to find matching host interface for IP %s", ip)
+	return ip, nil
 }
--- a/pkg/startup/startup_test.go
+++ b/pkg/startup/startup_test.go
@@ -41,6 +41,7 @@ import (
 	"github.com/projectcalico/libcalico-go/lib/names"
 	"github.com/projectcalico/libcalico-go/lib/net"
 	"github.com/projectcalico/libcalico-go/lib/options"
+	"github.com/projectcalico/node/pkg/startup/autodetection"
 )
 
 var exitCode int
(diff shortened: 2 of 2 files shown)


njuptlzf:
I can't find a way to request a review

caseydavenport:
@njuptlzf sorry for the delay on this - I seem to have missed it.

Assigning myself now.

pkg/startup/startup.go
```diff
@@ -1365,3 +1370,54 @@ func extractKubeadmCIDRs(kubeadmConfig *v1.ConfigMap) (string, string, error) {
...
+		if s[i] == c {
+			return i
+		}
+	}
+	return -1
+}
+
+// return the cidr of the dest on local
+func getLocalCidr(ip string, version int) (string, error) {
+	i := indexByteString(ip, '/')
```
caseydavenport, line 1387: Might be simpler to just check if the string has a `/` in it using the `strings` package:

```
if strings.Contains(ip, "/") {
  // Already a CIDR
  return ip, nil
}
```

Then we don't need to copy the `indexByteString` function here locally.

pkg/startup/startup.go
```diff
@@ -1365,3 +1370,54 @@ func extractKubeadmCIDRs(kubeadmConfig *v1.ConfigMap) (string, string, error) {
...
+	// Get a full list of interface and IPs and find the CIDR matching the
+	// found IP.
+	ifaces, err := autodetection.GetInterfaces(nil, nil, version)
+	if err != nil {
+		return ip, err
+	}
+	for _, iface := range ifaces {
+		log.WithField("Name", iface.Name).Info("Checking interface CIDRs")
+		for _, cidr := range iface.Cidrs {
+			log.WithField("CIDR", cidr.String()).Info("Checking CIDR")
```
caseydavenport, line 1403: I think we should demote both of these logs to `Debug` level.

pkg/startup/startup.go
```diff
@@ -1365,3 +1370,54 @@ func extractKubeadmCIDRs(kubeadmConfig *v1.ConfigMap) (string, string, error) {
...
+			var destCidr net.IP
```
caseydavenport, line 1404: I think we can parse the IP address once, outside of the loop, rather than do it every time again inside the loop.

pkg/startup/startup.go
```diff
@@ -1365,3 +1370,54 @@ func extractKubeadmCIDRs(kubeadmConfig *v1.ConfigMap) (string, string, error) {
...
+			if version == 4 {
+				destCidr = net.ParseIP(ip).To4()
+			} else {
+				destCidr = net.ParseIP(ip).To16()
+			}
+			if destCidr == nil {
+				return ip, fmt.Errorf("%s is unreasonable not a reasonable ip str", ip)
+			}
+			if cidr.IP.Equal(destCidr) {
+				log.WithField("CIDR", cidr.String()).Info("Found local CIDR ")
```
caseydavenport, line 1414: Can we tweak the wording of this log?

I'm thinking: "Including CIDR information from host interface"

pkg/startup/startup.go
```diff
@@ -1365,3 +1370,54 @@ func extractKubeadmCIDRs(kubeadmConfig *v1.ConfigMap) (string, string, error) {
...
+				return cidr.String(), nil
+			}
+		}
+	}
+
+	// Even if no cidr is found, it doesn't think it needs to throw an exception
+	log.Warnf("Unable to confirm IPv%d CIDR of %s", version, ip)
```
caseydavenport, line 1421: "Unable to find matching host interface for IP" might be a more accurate log.

pkg/startup/startup.go
```diff
@@ -1365,3 +1370,54 @@ func extractKubeadmCIDRs(kubeadmConfig *v1.ConfigMap) (string, string, error) {
...
+	for i := 0; i < len(s); i++ {
+		if s[i] == c {
+			return i
+		}
+	}
+	return -1
+}
+
+// return the cidr of the dest on local
+func getLocalCidr(ip string, version int) (string, error) {
```
caseydavenport, line 1386: nit: CIDR should be same-case in golang, since it is an acronym.

So, this should be `getLocalCIDR`

pkg/startup/startup.go
```diff
@@ -1365,3 +1370,54 @@ func extractKubeadmCIDRs(kubeadmConfig *v1.ConfigMap) (string, string, error) {
...
+func indexByteString(s string, c byte) int {
+	for i := 0; i < len(s); i++ {
+		if s[i] == c {
+			return i
+		}
+	}
+	return -1
+}
+
+// return the cidr of the dest on local
```
caseydavenport, line 1385: Let's enhance this comment a bit. I'm thinking something like this:

```
// getLocalCIDR attempts to merge CIDR information from the host with the given IP address.
// If a CIDR is provided, then it is simply returned.
// If an IP is provided, it attempts to find the matching interface on the host to detect the appropriate prefix length.
// If no match is found, the IP will be returned unmodified. 
```

pkg/startup/startup.go
```diff
@@ -535,7 +535,12 @@ func configureIPsAndSubnets(node *api.Node) (bool, error) {
 		validateIP(node.Spec.BGP.IPv4Address)
 	} else if ipv4Env != "none" {
 		if ipv4Env != "" {
-			node.Spec.BGP.IPv4Address = parseIPEnvironment("IP", ipv4Env, 4)
+			// Attempt to get the local cidr of ipv4Env
+			ipv4CidrOrIP, err := getLocalCidr(ipv4Env, 4)
```
caseydavenport, line 539: golang conventions say acronyms should be same-case, so this should be `ipv4CIDROrIP`

caseydavenport (changes requested):
Sorry again for the delay in spotting this!

I've added some comments. I think this is really cool. Thanks much for the PR!

njuptlzf:
Thanks for review, I will make changes

pkg/startup/startup.go
njuptlzf: fix

njuptlzf:
fixed, ptal @caseydavenport

pkg/startup/startup_test.go
```diff
@@ -1111,3 +1112,36 @@ var _ = Describe("UTs for monitor-addresses option", func() {
...
+		It("get the local cidr", func() {
+
+			cmd := exec.Command("ip", "addr", "add", "1.2.3.4/24", "dev", "lo")
+			if err := cmd.Run(); err != nil {
+				Skip("ip addr add failed:" + err.Error())
+			}
+
+			ipv4Env := "1.2.3.4"
+			ipv4CIDROrIP, _ := getLocalCIDR(ipv4Env, 4)
+			Expect(ipv4CIDROrIP).To(Equal("1.2.3.4/24"))
```
caseydavenport, line 1127: If this assertion fails then I think we'll fail to clean up the IP address that we added in the first step.

I think we probably want to include the cleanup in an `AfterEach()` or `defer()` so that it is always executed.

Alternatively, and likely preferably, we could mock out the `autodetection.GetInterfaces()` function so that we don't even need to interact with the host. Then we could have our fake "GetInterfaces()" return whatever interfaces we'd like for the tests without modifying the host's networking.

njuptlzf: Amazing!
 I will modify and retest

njuptlzf: fixed.
Constructing fake` autodetection.GetInterfaces`  is not feasible for built-in functions, so I use `defer()` . 
Do you think this test case needs to be deleted?

caseydavenport: I was imagining we could do something like changing the arguments to `getLocalCIDR`.

e.g.,

```
func getLocalCIDR(ip string, version in, getInterfaces func() ([]Interface, error)) (string, error) {}
```

Then, we can pass `autodetection.GetInterfaces` in the real code, and our own implementation in the test code.

njuptlzf: I understand what you mean. 
Yesterday I was also entangled in whether to pass in the interface as a parameter. This is why it took so long for me to update the submission.

Ok i will make changes again

njuptlzf: fixed,thanks.
ptal @caseydavenport

njuptlzf:
ptal @caseydavenport

caseydavenport (approved):
LGTM - Thanks!

caseydavenport:
@njuptlzf looks like it's failing our golang linter

```
pkg/startup/startup_test.go:1121:5: composites: `github.com/projectcalico/node/pkg/startup/autodetection.Interface` composite literal uses unkeyed fields (govet)01:36
				{"eth1",01:36
				^01:36
pkg/startup/startup_test.go:1138:5: composites: `github.com/projectcalico/node/pkg/startup/autodetection.Interface` composite literal uses unkeyed fields (govet)01:36
				{"eth1",01:36
				^
```

njuptlzf:
> @njuptlzf looks like it's failing our golang linter
> 
> ```
> pkg/startup/startup_test.go:1121:5: composites: `github.com/projectcalico/node/pkg/startup/autodetection.Interface` composite literal uses unkeyed fields (govet)01:36
> 				{"eth1",01:36
> 				^01:36
> pkg/startup/startup_test.go:1138:5: composites: `github.com/projectcalico/node/pkg/startup/autodetection.Interface` composite literal uses unkeyed fields (govet)01:36
> 				{"eth1",01:36
> 				^
> ```

Sorry, I don’t know how to test with vet first, so I used lazy initialization.
Tested with go vet, now fixed

champtar:
If you are looking for a workaround, see https://github.com/projectcalico/calico/issues/4254

PR #601, opened 2020-10-20
Reduce log level for spammy log

caseydavenport:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/startup/startup.go
+++ b/pkg/startup/startup.go
@@ -627,7 +627,7 @@ func validateIP(ipn string) {
 	for _, i := range ifaces {
 		for _, c := range i.Cidrs {
 			if ipAddr.Equal(c.IP) {
-				log.Infof("IPv%d address %s discovered on interface %s", ipAddr.Version(), ipAddr.String(), i.Name)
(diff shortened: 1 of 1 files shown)


pkg/startup/startup.go
```diff
@@ -627,7 +627,7 @@ func validateIP(ipn string) {
 	for _, i := range ifaces {
 		for _, c := range i.Cidrs {
 			if ipAddr.Equal(c.IP) {
-				log.Infof("IPv%d address %s discovered on interface %s", ipAddr.Version(), ipAddr.String(), i.Name)
+				log.Debugf("IPv%d address %s discovered on interface %s", ipAddr.Version(), ipAddr.String(), i.Name)
```
ozdanborne, line 630: A bit of a bummer that we won't see this message on clusters with default log levels set, especially considering ip autodetection is often a source of customer issues.

caseydavenport: I don't think this log is very important. It's the "mainline, working" branch, and is surrounded by higher level logs for the "unexpected, maybe not working" scenarios.

PR #742, opened 2020-12-23
Add test for service advertisement node exclusion

caseydavenport:
## Description

Add's a system test for the logic added in https://github.com/projectcalico/confd/pull/420/

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/tests/k8st/tests/test_bgp_advert.py
+++ b/tests/k8st/tests/test_bgp_advert.py
@@ -319,6 +319,129 @@ EOF
             # Assert that clusterIP is no longer an advertised route.
             retry_until_success(lambda: self.assertNotIn(local_svc_ip, self.get_routes()))
 
+    def test_node_exclusion(self):
+        """
+        Tests the node exclusion label.
+        - Create services, assert advertised from all nodes.
+        - Label one node so that it is excluded, assert that routes are withdrawn from that node.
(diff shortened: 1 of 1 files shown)


lwr20 (approved):
LGTM

tests/k8st/tests/test_bgp_advert.py
```diff
@@ -319,6 +319,129 @@ def test_cluster_ip_advertisement(self):
...
+
+            # Scale local service to 4 replicas
+            self.scale_deployment(local_svc, self.ns, 4)
+            self.wait_for_deployment(local_svc, self.ns)
+            self.wait_for_deployment(cluster_svc, self.ns)
+
+            # Assert routes are correct and services are accessible.
+            # Local service should only be advertised from nodes that can run pods.
+            # The cluster CIDR should be advertised from all nodes.
+            self.assert_ecmp_routes(local_svc_ip, [self.ips[1], self.ips[2], self.ips[3]])
```
mgleung, line 386: we don't need to check `self.ips[0]` for `local_svc_ip`?

caseydavenport: nope, self.ips[0] is the controller node, which doesn't get pods scheduled to it.

PR #751, opened 2021-01-05
Autodetect IP option to extract IP from kubernetes node status

robbrockbank:
## Description
Add new node auto detection method using kubernetes node status.

Not sure this really makes sense.... perhaps for k8s backend we should just always attempt to get the ip address from the k8s node status.  Not sure... but then also have to handle etcd backend, so maybe this is ok.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/startup/autodetection/autodetection_suite_test.go
+++ b/pkg/startup/autodetection/autodetection_suite_test.go
@@ -7,6 +7,7 @@ import (
 	"testing"
 
 	"github.com/onsi/ginkgo/reporters"
+
 	"github.com/projectcalico/libcalico-go/lib/testutils"
 )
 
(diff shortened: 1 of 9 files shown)


fasaxc:
@sridhartigera recently made a change (for BPF mode) to have `HostMetadataUpdate` fall back on the k8s node IP if the BGP IP isn't populated. https://github.com/projectcalico/libcalico-go/pull/1387

Might be some overlap?  I do like the idea of having an autodetection method that picks one of the k8s IPs though; should work well for environments like EKS

caseydavenport:
This is very conflicty now, and I think this might be better suited as an operator change.

For example, this can be implemented very easily via the Kubernetes downward API without any Calico code changes,

e.g.,

```yaml
- name: IP
  valueFrom:
    fieldRef:
      fieldPath: status.hostIP
```

gautvenk:
@caseydavenport can you please let me know if setting _**IP=status.hostIP**_ is being addressed at operator side?

PR #890, opened 2021-03-28
Fix winrm hang on Wait-ForCalicoInit

cdd-aix:
CalicoFelix service dependso n CalicoNode service, therefore we are
safe to remove Start-Service CalicoNode and Wait-ForCalicoInit
Plus,
https://github.com/projectcalico/node/blob/master/windows-packaging/CalicoWindows/felix/felix-service.ps1
waits on CalicoInit as well.

## Description

- bug fix
- Installing calico through 'vagrant winrm' and the vagrant shell provisioner hangs waiting for CalicoNode to start.  The wait is not needed as it's also called through the CalicoFelix wrapper script and CalicoFelix has a service dependency on CalicoNode
- Manually tested with https://github.com/cdd-aix/microk8s-kubernetes/tree/develop ... install-calico does not hang
- node/windows-packaging component is modified
- No issues opened before submitting this PR.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/windows-packaging/CalicoWindows/install-calico.ps1
+++ b/windows-packaging/CalicoWindows/install-calico.ps1
@@ -56,8 +56,6 @@ else
 Write-Host "Starting Calico..."
 Write-Host "This may take several seconds if the vSwitch needs to be created."
 
-Start-Service CalicoNode
-Wait-ForCalicoInit
 Start-Service CalicoFelix
 
 if ($env:CALICO_NETWORKING_BACKEND -EQ "windows-bgp")


song-jiang:
Thanks @cdd-aix for your PR.  I think calling `Start-Service CalicoNode and Wait-ForCalicoInit` in `install-calico.ps1` is still a good practise since most of Calico issues related to platforms are around NodeService. It is good to be able to log errors explicitly during installation period. 

Not sure how this would break vagrant winrm though.

cdd-aix:
I'm not sure how it's breaking winrm either.  I do know the Windows process
model provides too many corner-cases.  And wmi calls seem quite fragile.

The existing wait-forcalicoinit just shows time stamps... nothing to
indicate an error in the log.
I'd like to tackle a scenario where CalicoNode fails to start and provide
actionable information from the installer.  Can you point me to a
CalicoNode start failure ticket or few?

I want to fix the fails under winrm issue because ansible, vagrant, packer,
and terraform will tend to start the installation through winrm.

song-jiang:
I don't think we should make any change until we fully understand how winrm was broke by installation script. Please reopen the PR if you think otherwise.

PR #906, opened 2021-04-06
Bump libcalico and k8s lib version

rikatz:
This PR updates libcalico and k8s lib to the latest versions.


fasaxc:
Sorry, didn't see this and ended up adding it to #912 when I spotted that #912 hadn't gone through.

Issue #1062, opened 2021-07-08
make windows cni dir configurable


jayunit100:
install-calico.ps1 which calls config.ps1,
 seems to hardcode the way we find the CNI binary directory (i use C:/opt/cni/bin, but i think in your scripts, you assume containerd + cni are in the same place)
 
 Ive made a PR here https://github.com/kubernetes-sigs/sig-windows-dev-tools/pull/51 
 
 which almost gets calico working on our evironments, but we just need this tiny config change to finish the job :) 

thanks @song-jiang  !

song-jiang:
As discussed on PR 1063, we leave to users to modify config.ps1 but keep config.ps1 one-item-per-line and hold real values.

caseydavenport:
@song-jiang is there a way to configure this already then that's just not clearly documented? I think we need to support this. Sorry I don't know the windows install script well enough to answer on my own :sweat_smile:

PR #1063, opened 2021-07-09
Add non-mutating cni-conf-dir and cni-bin-dir defaulting to config.ps1

jayunit100:
a fix for #1062  (we can make install-calico.ps1 modular in a later PR if we want to)

### DIFF
--- a/windows-packaging/CalicoWindows/config.ps1
+++ b/windows-packaging/CalicoWindows/config.ps1
@@ -38,6 +38,34 @@ $env:ETCD_CA_CERT_FILE = "<your etcd ca cert>"
 ## CNI configuration, only used for the "vxlan" networking backends.
 
 # Place to install the CNI plugin to.  Should match kubelet's --cni-bin-dir.
+Write-Output "Checking CNI_BIN_DIR variable [ $env:CNI_BIN_DIR ] "
+
+if (-not (Test-Path env:CNI_BIN_DIR)) { 
+    Write-Output "Defaulting CNI_BIN_DIR..."
+
+    if (Get-IsContainerdRunning)
+    {
+        $env:CNI_BIN_DIR = Get-ContainerdCniBinDir
+    } else {
+        $env:CNI_BIN_DIR = "c:\k\cni" 
+    }
+}
+
+Write-Output "Checking CNI_CONF_DIR variable [ $env:CNI_CONF_DIR ] "
+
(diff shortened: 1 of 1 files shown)


song-jiang (changes requested):
Thanks for your PR @jayunit100 . 

The purpose of `config.ps1` is a single file which holds all configuration items for Calico Windows. If user want to update any item, they can easily do so by just replace the line for the item. e.g. $env:ETCD_CERT_FILE = "<your etcd cert>". 

Calico installation script is the example of that practise. On different cloud platforms, it replaces related configuration items. 

This one-config-one-line approach leaves config logic to the user. It also helps debugging as `config.ps` holds all the values.

song-jiang:
@lmm I just notices that there are extra line of `CNI-CONF_DIR` for containerd. Maybe we should move this out to Calico installation script in order to keep one-config-one-line?

jayunit100:
``` If user want to update any item, they can easily do so by just replacing the line for the item. e.g. $env:ETCD_CERT_FILE = "".``` 

really ? hmmm can we add an override functionality into this file to read from env vars so that we dont have to overwrite the whole thing ?

song-jiang:
I know it looks dumb to replace the whole string. But it makes config.ps1 simple to read and if user hit any issue, they can simply check the file to get all the values.

caseydavenport:
@song-jiang @jayunit100 what's the next step on this PR?

jayunit100:
i guess we should discuss at the community meeting ?   It seems to me like we should make this a json file if its not intended to be dynamic , because otherwise its tricky for end users to envision editing powershell as a configuration file, esp as we move to priveliged containers / pods running CNIs in 1.23+

song-jiang:
@jayunit100 Sorry I couldn't attend community meeting due to a conflict. As for your suggestion of using a json file, the fact is all Calico components (felix, node) reads Env variables for configurations. A json file means there are extra steps required (converting json to env) for each Calico component.

song-jiang:
With privileged container, we would set configurations through FELIX_CONFIG parameters for the daemonset hence config.ps1 could be removed completely.

caseydavenport:
Sounds like daemonset install with privileged containers support is the way forward - then we get k8s-native configuration through configmaps and CRDs more easily.

caseydavenport:
Going to close this for now. Acknowledge what is here isn't ideal, though.

PR #1064, opened 2021-07-12
windows: create VXLAN tunnel address in node

lmm:
## Description

Previously this was done in cni-plugin when the first pod is set up. But since then we create the IPAM block in node so we can create the tunnel address along with the network. Setting up the VXLAN tunnel address when setting up the network enables other approaches to installing Calico for Windows (e.g. with wins).

See also: https://github.com/projectcalico/cni-plugin/pull/1104

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
windows: create VXLAN tunnel address in node - previously this was done in cni-plugin when the first pod is set up
```

### DIFF
--- a/pkg/lifecycle/startup/startup_windows.go
+++ b/pkg/lifecycle/startup/startup_windows.go
@@ -20,6 +20,7 @@ import (
 	"net"
 	"os"
 	"strconv"
+	"time"
 
 	"github.com/sirupsen/logrus"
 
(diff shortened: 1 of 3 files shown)


pkg/lifecycle/startup/startup_windows.go
```diff
@@ -89,7 +90,17 @@ func ensureNetworkForOS(ctx context.Context, c client.Interface, nodeName string
 				return err
 			}
 			_, err = windows.SetupVxlanNetwork(networkName, subnet, uint64(vni), logrus.WithField("subnet", subnet.String()))
```
song-jiang, line 92: I suggest we add a `if err !=nil` here, so that we can catch error earlier if network is not created properly.

song-jiang (approved):
LGTM

PR #1101, opened 2021-07-28
Fix kube-proxy crashloopback for kind cluster

song-jiang:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


song-jiang:
I think CI is not passing because of calicoctl version mismatch change.

PR #1136, opened 2021-08-10
Update docker images and centos repos to address CVE false positives

coutinhop:
## Description

Update docker images and centos repos to address CVEs reported by Quay (advisories RHSA-2021:2717 and RHSA-2021:1611). These were false positives, as calico does not use systemd, but updating should stop the reporting.

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Dockerfile.amd64
+++ b/Dockerfile.amd64
@@ -13,7 +13,7 @@
 # limitations under the License.
 ARG ARCH=x86_64
 ARG GIT_VERSION=unknown
-ARG IPTABLES_VER=1.8.4-15
+ARG IPTABLES_VER=1.8.4-17
 ARG LIBNFTNL_VER=1.1.5-4
 ARG RUNIT_VER=2.1.2
 ARG BIRD_IMAGE=calico/bird:latest
@@ -31,7 +31,7 @@ ARG ARCH
 ARG IPTABLES_VER
 ARG LIBNFTNL_VER
 ARG RUNIT_VER
-ARG CENTOS_MIRROR_BASE_URL=http://vault.centos.org/8.3.2011
+ARG CENTOS_MIRROR_BASE_URL=https://vault.centos.org/8.4.2105
 ARG LIBNFTNL_SOURCERPM_URL=${CENTOS_MIRROR_BASE_URL}/BaseOS/Source/SPackages/libnftnl-${LIBNFTNL_VER}.el8.src.rpm
 ARG IPTABLES_SOURCERPM_URL=${CENTOS_MIRROR_BASE_URL}/BaseOS/Source/SPackages/iptables-${IPTABLES_VER}.el8.src.rpm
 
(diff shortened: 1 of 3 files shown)


Dockerfile.amd64
```diff
@@ -44,7 +44,9 @@ RUN dnf install -y 'dnf-command(config-manager)' && \
     # Need these to build runit.
     yum install -y wget glibc-static gcc && \
     # Ensure security updates are installed.
-    yum -y update-minimal --security --sec-severity=Important --sec-severity=Critical
+    yum -y update-minimal --security --sec-severity=Important --sec-severity=Critical && \
+    # Update systemd related packages due to CVEs (advisories RHSA-2021:2717 and RHSA-2021:1611)
+    yum -y update-minimal --advisory=RHSA-2021:2717 --advisory=RHSA-2021:1611
```
lmm, line 49: I think the CVEs are showing up because we’re using old versions (8.3) of ubi and centos to build our node image. You can see this at the scan results for v3.20.0: https://quay.io/repository/calico/node/manifest/sha256:913955a36179a53d36a0df7f26319a5278991eea2514eab339cce15df24c220c?tab=vulnerabilities The fix versions for these two libs is `0:239-45.el8_4.2` -- note the `8_4` -- so I think if we just update ubi and centos from 8.3 -> 8.4 that should remove the clear up the vuln scan.

lmm: We can verify whether the changes work by building an image from this branch and pushing it to quay.io

coutinhop: thanks for catching this, @lmm! the packages were being downgraded further along and the advisories were still showing up on Quay. The correct way to do this is to update the ubi image and centos repos to 8.4, which I just pushed.

centos.repo
```diff
@@ -1,13 +1,13 @@
...
-baseurl = https://vault.centos.org/8.3.2011/BaseOS/x86_64/os/
+baseurl = http://mirror.centos.org/centos/8.4.2105/BaseOS/x86_64/os/
 enabled = 1
 gpgkey = https://www.centos.org/keys/RPM-GPG-KEY-CentOS-Official
 gpgcheck = 1
 
 [centos-8-appstream]
 name = CentOS - AppStream
-baseurl = https://vault.centos.org/8.3.2011/AppStream/x86_64/os/
+baseurl = http://mirror.centos.org/centos/8.4.2105/AppStream/x86_64/os/
```
lmm, line 10: I suppose when Centos 8.4 is deprecated we'll need to switch this back to mirror.centos.org

lmm: Also weird that mirror.centos.org isn't served over TLS

lmm (approved):
Great!

centos.repo
lwr20: That's a security issue.  We can't trust the server's identity over HTTP, and that leave us vulnerable to a supply chain attack.  We can't allow a non TLS server.

lmm: The packages' integrity is checked with the Centos gpg key so we can trust the packages.

Server identity is another matter though. I suppose we could pick a mirror from: https://www.centos.org/download/mirrors/

coutinhop: Got it, I'll update the files to use https mirrors. Thanks!

PR #1166, opened 2021-08-19
Configurable VXLAN adapter

knabben:
## Description

This PR enables the capability to configure the adapter name when creating a new HNS network, this is useful when multiple NIC exist in the host.

When the value is not set the primary NIC should be used.

## Release Note

```release-note
Windows VXLAN network adapter can be configurable when multi interfaces exist.
```

### DIFF
--- a/windows-packaging/CalicoWindows/config.ps1
+++ b/windows-packaging/CalicoWindows/config.ps1
@@ -59,7 +59,8 @@ $env:CNI_IPAM_TYPE = "calico-ipam"
 $env:VXLAN_VNI = "4096"
 # Prefix used when generating MAC addresses for virtual NICs.
 $env:VXLAN_MAC_PREFIX = "0E-2A"
-
+# Network Adapter used on VXLAN, leave blank for primary NIC. 
+$env:VXLAN_ADAPTER = ""
 
 ## Node configuration.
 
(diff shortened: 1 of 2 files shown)


windows-packaging/CalicoWindows/config.ps1
```diff
@@ -59,7 +59,8 @@ $env:CNI_IPAM_TYPE = "calico-ipam"
 $env:VXLAN_VNI = "4096"
 # Prefix used when generating MAC addresses for virtual NICs.
 $env:VXLAN_MAC_PREFIX = "0E-2A"
-
+# Network Adapter used on VXLAN, leave blank for primary NIC. 
+$env:VXLAN_ADAPTER = ""
```
fasaxc, line 63: Does the default `""` value do something sensible (i.e. give us the existing behaviour)?

knabben: Yes, `New-HNSNetwork` checks for the value, if it does not exists use the primary interface, what is the default value

https://github.com/microsoft/SDN/blob/b189fc96bcba8ea3ed6c77b51bca9266a8683ae7/Kubernetes/windows/hns.psm1#L208

knabben: Yes, `New-HNSNetwork` checks for the value - https://github.com/microsoft/SDN/blob/b189fc96bcba8ea3ed6c77b51bca9266a8683ae7/Kubernetes/windows/hns.psm1#L208

If the value does not exist, the primary interface is used, giving us the current behavior.

song-jiang (approved):
LGTM

knabben:
> @song-jiang @knabben is there a corresponding docs PR for this?

No, I will open a new one adding this to the docs.

@caseydavenport what section does this parameter documentation go?

PR #1201, opened 2021-09-02
Add Windows docker image for calico node

jsturtevant:
## Description

I've trying out different cni's with Windows HostProcess containers and started to try out calico. Initially was working through powershell scripts but feel like using the same type of approach Linux uses is better.

This is a rough draft and can be combined with https://github.com/projectcalico/cni-plugin/pull/1148 to run calico in Hostprocess containers.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Adds support for Windows HostProcess containers
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -8,3 +8,6 @@
 # - CentOS repository file and our license
 !centos.repo
 !LICENSE
+!windows-packaging/CalicoWindows/libs
+!windows-packaging/CalicoWindows/node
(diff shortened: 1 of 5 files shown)


jsturtevant:
The yaml I've used to test and deploy this is https://gist.github.com/jsturtevant/523368ba72abbbc253d13433b7eb32d6. I wasn't sure where in the calico repositories this should live.

PR #1202, opened 2021-09-02
Bump BIRD, to treat 240.0.0.0/4 IPv4 range like RFC 1918 ranges

neiljerram:


Makefile
```diff
@@ -456,7 +456,7 @@ remove-go-build-image:
 
 .PHONY: st
 ## Run the system tests
-st: image-all remote-deps dist/calicoctl busybox.tar calico-node.tar workload.tar run-etcd calico_test.created dist/calico dist/calico-ipam
+st: image remote-deps dist/calicoctl busybox.tar calico-node.tar workload.tar run-etcd calico_test.created dist/calico dist/calico-ipam
```
neiljerram, line 459: @Brian-McM Could you review this change, as I think you added the `image-all` here, as part of a commit entitled "Refactor semaphore CI so it runs faster"

caseydavenport: I believe we still need semaphore to at least cross-build images as part of CI to ensure we don't break the build for alternative architectures.

Maybe we need to have a target specifically for that, though - e.g., one that builds and deletes each image in sequence so as to avoid filling up the disk?

PR #1205, opened 2021-09-03
Add 'make image-all' to CI

coutinhop:
Add 'make image-all' target to Makefile and corresponding job to semaphore.yml. Also add a job to build just the amd64 image (which is used by STs) as a prereq to run STs.

## Todos
- [X] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -28,6 +28,16 @@ global_job_config:
       - git fetch --unshallow
 
 blocks:
+  - name: "Build Images"
+    dependencies: []
+    task:
+      jobs:
+        - name: "Build All Images"
+          execution_time_limit:
+            minutes: 120
+          commands:
+            - TEST_IMAGE_BUILD=true make image-all
+
   - name: "General Tests"
     dependencies: []
     task:
@@ -40,8 +50,18 @@ blocks:
             - make ut
             - make fv
 
-  - name: "System Tests"
+  - name: "Build amd64 Image"
     dependencies: []
+    task:
+      jobs:
+        - name: "Build amd64 Image"
+          execution_time_limit:
+            minutes: 120
+          commands:
+            - make image
+
+  - name: "System Tests"
+    dependencies: ["Build amd64 Image"]
     task:
       prologue:
         commands:
--- a/Makefile
+++ b/Makefile
@@ -255,6 +255,17 @@ image: remote-deps $(NODE_IMAGE)
 image-all: $(addprefix sub-image-,$(VALIDARCHES))
 sub-image-%:
 	$(MAKE) image ARCH=$*
+ifeq ($(TEST_IMAGE_BUILD),true)
+	# If testing image builds, clean sub-image afterwards to free disk space (for Semaphore CI)
+	$(MAKE) clean-sub-image-$*
+endif
+
+## Remove images for all supported ARCHes
+clean-image-all: $(addprefix clean-sub-image-,$(VALIDARCHES))
+## Remove sub-image from docker and delete $(NODE_CONTAINER_CREATED) file
+clean-sub-image-%:
+	rm -f .calico_node.created-$*
+	docker rmi $(NODE_IMAGE):latest-$* || true
 
 $(NODE_IMAGE): $(NODE_CONTAINER_CREATED)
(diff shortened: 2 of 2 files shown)


frozenprocess (commented):
/LGTM

Makefile
```diff
@@ -256,6 +256,18 @@ image-all: $(addprefix sub-image-,$(VALIDARCHES))
 sub-image-%:
 	$(MAKE) image ARCH=$*
 
+## Test building images for all supported ARCHes
+test-image-all: $(addprefix test-sub-image-,$(VALIDARCHES))
```
Brian-McM, line 260: Why do we need a "test-image-all", is there something wrong with "image-all"?

Brian-McM (commented):
If we're adding a build step I'd say it should be it's own block, and should be a pre req for all the tests (except the static checks). You could possibly add some UT tests to that block too (like typha does), but I think a quick build test before running any tests is useful, because we won't waste any of our semaphore quota on big tests if the build fails.

You may even want to consider caching the image created for the next set of blocks so they don't waste time building an image that's already been built, but I'd say that could be follow up work (not always worth it to cache).

Makefile
coutinhop: @caseydavenport had mentioned that storage might fill up on semaphore from building all images, so "test-image-all" and "test-sub-image-<arch>" are there just to clean up after building each image.

coutinhop:
> If we're adding a build step I'd say it should be it's own block, and should be a pre req for all the tests (except the static checks). You could possibly add some UT tests to that block too (like typha does), but I think a quick build test before running any tests is useful, because we won't waste any of our semaphore quota on big tests if the build fails.
> 
> You may even want to consider caching the image created for the next set of blocks so they don't waste time building an image that's already been built, but I'd say that could be follow up work (not always worth it to cache).

I'm not sure I follow, sorry. 'image-all' was removed as a requirement from the 'st' target, as the tests only run on the amd64 arch, so there wasn't a need to build all images for the STs (several times, because we split and parallelize parts of them). But then we should still have a CI job building all arch's to make sure nothing breaks any of them. Since we are still building amd64 to run the STs and building all images takes longer than the tests to run (about 1 hour on the runs for this PR), I'm not sure if making this a prereq would be best.

Maybe just the amd64 as a prereq and caching it is good, but I'm not sure if it's worth the effort, since it's relatively quick (about 5 min from looking at some of the job logs).

Makefile
Brian-McM: That's understandable, but what we really want to test is image-all, if that target changes we need to now update this target as well.

I'd say test the image-all target, since that's the goal, and if you need to clean up then just clean up the images, as opposed to redefining image-all and it's sub images.

Brian-McM:
> I'm not sure I follow, sorry. 'image-all' was removed as a requirement from the 'st' target, as the tests only run on the amd64 arch, so there wasn't a need to build all images for the STs (several times, because we split and parallelize parts of them). But then we should still have a CI job building all arch's to make sure nothing breaks any of them. Since we are still building amd64 to run the STs and building all images takes longer than the tests to run (about 1 hour on the runs for this PR), I'm not sure if making this a prereq would be best.

So there were two parts to my comment 1) testing the build works before we spend time running the tests and 2) caching the whatever images we build so we don't need to cache later. 

1) should be done as a pre req, not for the test _target_ but for the the _semaphore blocks_ (look at how we have it for felix, and almost every other project). This makes it so that the 10 other blocks (each which take up an instance) won't execute if the build fails, which is ideal.

2) would be a nice to have, and you don't need to cache all the images if we're not using them, just whatever else we need.

Makefile
coutinhop: Thanks @Brian-McM! I changed the 'sub-image-%' target to conditionally clean based on an env var and am using it on the CI job.

.semaphore/semaphore.yml
```diff
@@ -28,6 +28,16 @@ global_job_config:
...
 blocks:
+  - name: "Build Images"
+    dependencies: []
+    task:
+      jobs:
+        - name: "Build All Images"
+          execution_time_limit:
+            minutes: 120
+          commands:
+            - TEST_IMAGE_BUILD=true make image-all
```
mgleung, line 39: I think you should be able to set `TEST_IMAGE_BUILD` in an `env` section in this block somewhere right?

coutinhop: Right, yeah. Will do, thanks for pointing that out!

PR #1224, opened 2021-09-27
Run Calico node as non-root user

mgleung:
## Description

Set the SUID bits on the binaries that require root permissions and change directory permissions to the root group to allow for node to run as a non-root user.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Change node and supporting binary permissions so that they can be run as a non-root user
```


lmm (approved):
LGTM - we chatted about this over Slack.

PR #1226, opened 2021-09-27
Update Kubernetes version used in tests

caseydavenport:
## Description

- Update to use upstream kind rather than our fork
- Update version of k8s
- Update Calico manifests
- Update test code to be compatible with new k8s version

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


Makefile
```diff
@@ -83,7 +83,8 @@ ifneq ($(BUILDARCH),amd64)
 	ETCD_IMAGE=$(ETCD_IMAGE)-$(ARCH)
 endif
 
-HYPERKUBE_IMAGE?=gcr.io/google_containers/hyperkube-$(ARCH):$(K8S_VERSION)
+# TODO: Update this to use newer version of Kubernetes.
```
caseydavenport, line 86: I still need to fix this for the couple of tests that use it.

mgleung (approved):
LGTM

Makefile
caseydavenport: Going to fix this in a subsequent PR

PR #1234, opened 2021-10-01
Added node status reporter

song-jiang:
## Description

This PR added support for calico-node to report back node status (bird status, bgp peers , routes etc.) via CalicoNodeStatus API. 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


Makefile
```diff
@@ -225,6 +225,9 @@ DOCKER_GO_BUILD_CGO=$(DOCKER_RUN) -e CGO_ENABLED=$(CGO_ENABLED) $(CALICO_BUILD)
 $(NODE_CONTAINER_BINARY): $(LOCAL_BUILD_DEP) $(SRC_FILES) go.mod
 	$(DOCKER_GO_BUILD_CGO) sh -c '$(GIT_CONFIG_SSH) go build -v -o $@ $(BUILD_FLAGS) $(LDFLAGS) ./cmd/calico-node/main.go'
 
+dist/status: $(LOCAL_BUILD_DEP) $(SRC_FILES) go.mod
```
song-jiang, line 228: This is just for debug purpose. Will get it removed later.

cmd/calico-node/main.go
```diff
@@ -65,6 +66,10 @@ var felixReady = flagSet.Bool("felix-ready", false, "Run felix readiness checks"
 // thresholdTime is introduced for bird readiness check. Default value is 30 sec.
 var thresholdTime = flagSet.Duration("threshold-time", 30*time.Second, "Threshold time for bird readiness")
 
+// Options for node status.
+var runStatusReporter = flagSet.Bool("status-reporter", false, "Run node status reporter")
+var showStatus = flagSet.Bool("show-status", false, "Print out node status")
```
song-jiang, line 71: `show-status` would allow somebody to quickly checking node status by 
```
kubectl exec -t calico-node-xxx -n calico-system -- calico-node -show-status
```

cmd/status/main.go
```diff
@@ -0,0 +1,9 @@
+package main
+
+import (
+	"github.com/projectcalico/node/pkg/status"
+)
+
+func main() {
+	status.Show()
+}
```
song-jiang, line 9: This is just for debug purpose. Will get it removed later.

pkg/status/nodestatus_test.go
```diff
@@ -0,0 +1,1059 @@
+// Copyright (c) 2018,2021 Tigera, Inc. All rights reserved.
```
song-jiang, line 1: This is not done yet.

pkg/status/reporter.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+}
+
+// reportStatus queries Bird or other components and update node status resource.
+func (r reporter) reportStatus() {
+	var err error
+	// The idea here is that we either update everything successfully or we update nothing.
+
+	// Make a local copy first.
+	status := *r.status
```
song-jiang, line 154: Just before we report back status, do we need to read the latest version of it? But the things is only this node is updating the resource so maybe we don't need that?

pkg/status/bird_status.go
```diff
@@ -0,0 +1,220 @@
...
+	ready            bool
+	version          string
+	routeID          string
+	serverTime       string
+	lastBootTime     string
+	lastReconfigTime string
+}
+
+func (b *birdStatus) toNodeStatusAPI() apiv3.CalicoNodeBirdStatus {
+	return apiv3.CalicoNodeBirdStatus{
```
caseydavenport, line 44: Might be simpler if instead of duplicating fields on the `birdStatus` struct, we just had a single field like this:

```go
type birdStatus struct {
  status apiv3.CalicoNodeBirdStatus
}
```

Then this function becomes effectively unnecessary:

```go
func (b *birdStatus) toNodeStatusAPI() apiv3.CalicoNodeBirdStatus {
  return b.status
}
```

pkg/status/bird_status.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+}
+
+// Unmarshal from a line in the BIRD protocol output.  Returns true if
+// successful, false otherwise.
+func (b *birdStatus) unmarshalBIRD(line string) bool {
+	// Peer names will be of the format described by bgpPeerRegex.
+	log.Debugf("Parsing line: %s", line)
+
+	if strings.Contains(line, "BIRD") && strings.Contains(line, "ready") {
```
caseydavenport, line 60: We should make sure to have extensive tests in case BIRD every updates / changes its output format, since this logic is highly dependent on that.

pkg/status/peers.go
```diff
@@ -0,0 +1,312 @@
...
+		Type:   bgpTypeMap[b.peerType],
+		State:  b.state,
+		Since:  b.since,
+		Reason: info,
+	}
+}
+
+// Unmarshal a peer from a line in the BIRD protocol output.  Returns true if
+// successful, false otherwise.
+func (b *bgpPeer) unmarshalBIRD(line, ipSep string) bool {
```
caseydavenport, line 77: Any reason to use a `bool` here rather than an `error`? 

It seems to be that we should always expect this to pass unless there is a bug in our code or a real error in BIRD generating bad output, and we should be alerting the user to that.

I think as it is right now, we just log at debug level for errors, which might not be enough.

pkg/status/reporter.go
```diff
@@ -0,0 +1,209 @@
...
+	// channel to indicate this reporter is not needed anymore.
+	// It should start termination process.
+	done chan struct{}
+
+	// channel to indicate this reporter is terminated.
+	term chan struct{}
+}
+
+// newReporter creates a reporter and start running a goroutine handling resource update.
+func newReporter(name string, client client.Interface, intervalInSeconds int, populators map[BirdConnType]map[apiv3.NodeStatusClassType]statusPopulator) *reporter {
```
caseydavenport, line 71: I think maybe we should have `newReporter` take a NodeStatus object so that we can seed it with initial configuration before calling `run()`

I think it simplifies a bit of the logic, since we'll never have a reporter without a `status` set, which is an invalid state anyway really (since reporters only exist because of a status object)

pkg/status/reporter.go
```diff
@@ -0,0 +1,209 @@
...
+// run is the main reporting loop, it loops until done.
+func (r reporter) run() {
+	runImmediately := make(chan struct{})
+
+	for {
+		select {
+		case latest := <-r.ch:
+			// Received an update of node status resource.
+			if latest.Name != r.name {
+				log.Warningf("node status reporter (%s) receive request with different name (%s), ignore it", r.name, latest.Name)
```
caseydavenport, line 126: This is really a bug in the code if we hit this, right? We can probably make this louder (error level).

I'd also recommend simplifying the conditional logic here by putting a `continue` in this branch and removing the `else`, which makes it more clear that this is not the mainline code path.

pkg/status/reporter.go
```diff
@@ -0,0 +1,209 @@
...
+			} else {
+				r.status = latest
+				r.checkAndUpdateTicker(latest.Spec.UpdateIntervalInSeconds)
+				// kick start node status update
+				runImmediately <- struct{}{}
+			}
+		case <-runImmediately:
+		case <-r.ticker.C:
+			if r.status == nil {
```
caseydavenport, line 135: if we initialize the reporter with a status object, this check can go away

pkg/status/reporter.go
caseydavenport: We definitely need to handle resource conflicts, even if it should be rare. 

A user could modify the node status resource, for example.

pkg/status/reporter.go
```diff
@@ -0,0 +1,209 @@
...
+	timeout := time.After(3 * time.Second)
+	for {
+		select {
+		case <-timeout:
+			err = fmt.Errorf("timed out patching node status, last error was: %s", err.Error())
+			log.WithError(err).Warn("failed to report node status")
+			return
+		default:
+			status.Status.LastUpdated = metav1.Time{Time: time.Now()}
+			_, err = r.client.CalicoNodeStatus().Update(context.Background(), &status, options.SetOptions{})
```
caseydavenport, line 199: This should probably be `UpdateStatus()` really.

I don't think we have status on most of our Calico APIs at the moment, but we really don't want this controller to accidentally overwrite a user's spec changes when it tries to set a status update.

At a minimum we need to handle conflict updates gracefully (I suspect we do eventually handle a conflict error with a retry, since the syncer will get the event and eventually send it through to here.

pkg/status/reporter.go
```diff
@@ -0,0 +1,209 @@
...
+	if reflect.DeepEqual(status, *r.status) {
+		// Nothing has changes since last time we updated.
+		return
+	}
+
+	// Update resource
+	timeout := time.After(3 * time.Second)
+	for {
+		select {
+		case <-timeout:
```
caseydavenport, line 193: You should be able to get rid of this branch and instead just use a `context.WithTimeout` for the context passed to `Update()`

pkg/status/reporter.go
```diff
@@ -0,0 +1,209 @@
...
+			log.WithError(err).Warn("failed to report node status")
+			return
+		default:
+			status.Status.LastUpdated = metav1.Time{Time: time.Now()}
+			_, err = r.client.CalicoNodeStatus().Update(context.Background(), &status, options.SetOptions{})
+			if err != nil {
+				log.WithError(err).Warnf("Failed to update node status resource; will retry")
+			} else {
+				// Success!
+				r.status = &status
```
caseydavenport, line 204: I don't think `Update()` updates the pointed passed in. It does however return the updated object, so you should probably use that here instead.

e.g.,

```go
updated, err := client.Update(....)
```
and then
```go
r.status = updated
```

pkg/status/reporter.go
```diff
@@ -0,0 +1,209 @@
...
+
+	// Make a local copy first.
+	status := *r.status
+
+	needUpdate := false
+	for _, ipv := range []BirdConnType{BirdConnTypeV4, BirdConnTypeV6} {
+		failed := false
+		// Populate status from registered populators.
+		for _, class := range r.status.Spec.Classes {
+			if p, ok := r.populators[ipv][class]; ok {
```
caseydavenport, line 161: I'd recommend simplifying the branch structure here as well - make the first branch the `!ok` version and then the rest of the logic doesn't need to be in a branch.

I'm not 100% sure, but I think I'd also be in favor of failing the entire operation if one of the populators fails. I'm not sure what scenarios you were imagining, but I think it would be confusing for users to see an inconsistent set of data (e.g., updated, routes, but not updated peers).

```go
p, ok := r.populators[ipv][class]; 
if !ok {
  log.Warningf("Wrong class (%s) requested for node status reporter (%s)", class, r.name)
  continue
}

err = p.Populate(&status)
if err != nil {
  return err
}
```

pkg/status/reporter.go
```diff
@@ -0,0 +1,209 @@
...
+				log.Warningf("Wrong class (%s) requested for node status reporter (%s)", class, r.name)
+			}
+		}
+
+		if !failed {
+			needUpdate = true
+		}
+	}
+
+	if !needUpdate {
```
caseydavenport, line 179: See my suggestion above about simply erroring if any of the populators fail - I think that is my preferred approach rather than handling partial failures, since inconsistent state in the API will be confusing for users. 

However, if we do for some reason need to keep this, can't we just remove `needUpdate` and turn this check into `if !failed`?

pkg/status/route.go
```diff
@@ -0,0 +1,299 @@
...
+
+var (
+	viaRegex       = regexp.MustCompile(`(?P<dest>\S*)\s*via\s(?P<gateway>\S*)\s*on\s(?P<iface>.*)\s\[(?P<from>\S*)\s.*\].*`)
+	devRegex       = regexp.MustCompile(`(?P<dest>\S*)\s*dev\s(?P<iface>.*)\s\[(?P<from>\S*)\s.*\].*`)
+	blackholeRegex = regexp.MustCompile(`(?P<dest>\S*)\s*blackhole\s\[(?P<from>\S*)\s.*\].*`)
+)
+
+// route is a structure containing details about a route.
+type route struct {
+	dest        string
```
caseydavenport, line 42: same comment here - seems like we're duplicating information when we could just be building an `apiv3.CalicoNodeRoute` in place. WDYT?

pkg/status/route.go
```diff
@@ -0,0 +1,299 @@
+// Copyright (c) 2021 Tigera, Inc. All rights reserved.
```
caseydavenport, line 1: For clarity, it might be worth organizing all of the populators into a sub-package. 

e.g.,

```
pkg/status
pkg/status/populators
```

caseydavenport (changes requested):
Very quick pass at this - need to come back and take a more detailed look, but wanted to post what I had first.

pkg/status/route.go
song-jiang: Good point.

pkg/status/bird_status.go
song-jiang: I was thinking populators (in this case `birdStatus`) always has it's own internal structure and may convert to different API structures in the future. 

Say we may have another v3 API which holds part of the information from birdStatus. 

Another benefit of decoupling internal structure with API is that `show()` function may show more fields than the API does. WDYT?

pkg/status/nodestatus.go
```diff
@@ -0,0 +1,265 @@
...
+	// Read Typha settings from the environment.
+	// When Typha is in use, there will already be variables prefixed with FELIX_, so it's
+	// convenient if we honor those as well as the CALICO variables.
+	typhaConfig := syncclientutils.ReadTyphaConfig([]string{"FELIX_", "CALICO_"})
+	if syncclientutils.MustStartSyncerClientIfTyphaConfigured(
+		&typhaConfig, syncproto.SyncerTypeNodeStatus,
+		buildinfo.GitVersion, nodename, fmt.Sprintf("node-status %s", buildinfo.GitVersion),
+		r,
+	) {
+		log.Debug("Using typha syncclient")
```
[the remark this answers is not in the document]
song-jiang: I've got PR on typha repo https://github.com/projectcalico/typha/pull/662

pkg/status/peers.go
song-jiang: This is inherited from calicoctl. I think the idea is we could skip some lines which does not provide any information. It is not an error, it just means we don't need to worry about those lines. 

I'll update the code to make warning level for errors we do care about.

pkg/status/bird_status.go
song-jiang: Agree. I'm thinking to run tests with Node ST on kind. Couple of thoughts now.
**Infra**
- v4 cluster 
- Dual stack cluster
- Dual tor cluster

**BGP Peer type**
- NodeMesh
- BGPPeer

**Possible test cases**
- Node restart
- external BGP peer restart
- Pod restart (trigger a kernel route update)
- Kill bird or bird6 processes

pkg/status/reporter.go
song-jiang: Good point.

pkg/status/route.go
song-jiang: See my comment above. I still see value of having a internal structure.

pkg/status/reporter.go
song-jiang: The only concern I have is that if a cloud platform does not support ipv6 and we are constantly getting errors of that. We will never get update for ipv4. But maybe that is not the case, bird6 works fine even ipv6 is not supported by underlaying infra. We need do some test on it to be sure.

pkg/status/reporter.go
song-jiang: Nice. Thanks!

pkg/status/reporter.go
song-jiang: Let me simplify the code that we fail on any error. If we see issues regarding bird6, we can come back and fix it.

song-jiang: Normally I'm keen on the code structure you outlined above where only error handling is on a branch. Not sure why I didn't do it this time. 😄

pkg/status/bird_status.go
lwr20: Could also use policy to block TCP port 179 on one or more nodes to simulate network failure?

pkg/status/bird_status.go
caseydavenport: Maybe, seems like we don't have a clear need for that now which is why I was suggesting it. Doesn't seem worth it to structure the code around a hypothetical that we don't know we really will ever need.

pkg/status/nodestatus.go
```diff
@@ -0,0 +1,265 @@
...
+		// resource name is unique.
+		if v, ok := u.Key.(model.ResourceKey); ok {
+			name = v.Name
+		} else {
+			log.Warningf("Unexpected resource update: %s", u.Key)
+			continue
+		}
+
+		switch u.UpdateType {
+		case bapi.UpdateTypeKVDeleted:
```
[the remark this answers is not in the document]
caseydavenport: this one still needs fixing

pkg/status/bird_status.go
lwr20: I'm sure they could be configured away for this test :)

lwr20: If not, perhaps `tc qdisc add dev eth0 root netem loss 100%`.

pkg/status/reporter.go
```diff
@@ -0,0 +1,230 @@
...
+			if latest.Name != r.name {
+				r.logCtx.Errorf("node status reporter receive request with different name (%s), ignore it", latest.Name)
+				break
+			}
+			r.status = latest
+			r.checkAndUpdateTicker(latest.Spec.UpdateIntervalInSeconds)
+			// kick start node status update
+			runImmediately <- struct{}{}
+
+		case <-runImmediately:
```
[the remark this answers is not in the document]
song-jiang: Actually, `fallthrough` is supported in `select`. I think without it, if next time round we get a new update before reporting current one, it is fine.

pkg/status/nodestatus.go
```diff
@@ -0,0 +1,300 @@
...
+				continue
+			}
+		}
+	}
+}
+
+// processPendingUpdates processes pending updates in main loop.
+// It is called when we are in-sync.
+func (r *nodeStatusReporter) processPendingUpdates() {
+	if len(r.pendingUpdates) == 0 {
```
caseydavenport, line 263: This is redundant code - if pendingUpdates is empty, the loop below will just be skipped.

pkg/status/populators/bird_conn.go
```diff
@@ -0,0 +1,57 @@
...
+// getBirdConn return a connection to bird socket.
+func getBirdConn(ipv IPFamily) (*birdConn, error) {
+	birdSuffix := ipv.BirdSuffix()
+
+	// Try connecting to the bird socket in `/var/run/calico/` first to get the data
+	c, err := net.Dial("unix", fmt.Sprintf("/var/run/calico/bird%s.ctl", birdSuffix))
+	if err != nil {
+		// If that fails, try connecting to bird socket in `/var/run/bird` (which is the
+		// default socket location for bird install) for non-containerized installs
+		log.Debugln("Failed to connect to BIRD socket in /var/run/calic, trying /var/run/bird")
```
caseydavenport, line 49: typo - "calic"

caseydavenport: Do we need to handle non-containerized installs? Doesn't this code only ever run inside the calico/node container?

pkg/status/reporter.go
```diff
@@ -0,0 +1,248 @@
...
+	if reflect.DeepEqual(status.Status, r.status.Status) {
+		// Nothing has changes since last time we updated.
+		return nil
+	}
+
+	var err error
+	var updatedResource *apiv3.CalicoNodeStatus
+	// Update resource
+	for i := 0; i < 3; i++ {
+		ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
```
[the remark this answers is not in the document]
caseydavenport: I suppose this has a problem if the update interval set by the user is smaller than our timeout!

We might want to consider not retrying if the update interval is suitably small - say less than 30 seconds - since we're going to retry again shortly anyway.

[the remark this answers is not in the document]
song-jiang: Good point!

pkg/status/reporter.go
```diff
@@ -0,0 +1,261 @@
...
+
+package status
+
+import (
+	"context"
+	"math"
+	"reflect"
+	"time"
+
+	cerrors "github.com/projectcalico/libcalico-go/lib/errors"
```
caseydavenport, line 23: these don't need to be spaced out like this - is `goimports` doing this?

PR #1238, opened 2021-10-05
Add handling for setting up hostpath directories for non-root use

mgleung:
## Description

Adds hostpath handling so that we will not need a separate image in order to do the hostpath permission setup for non-root calico-node usage. This should be used similar to the following manifest:
```
        - name: hostpath-init
          image: docker.io/calico/node:non-root
          command: ['sh', '-c', 'calico-node -hostpath-init']
          env:
            - name: NODE_USER_ID
              value: "1000"
          securityContext:
            runAsUser: 0
          volumeMounts:
          - mountPath: /var/run
            name: var-run
            readOnly: false
          - mountPath: /var/lib
            name: var-lib
            readOnly: false
```

These changes will be required for https://github.com/tigera/operator/pull/1544 to install it properly

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/cmd/calico-node/main.go
+++ b/cmd/calico-node/main.go
@@ -33,6 +33,7 @@ import (
 	"github.com/projectcalico/node/pkg/allocateip"
 	"github.com/projectcalico/node/pkg/cni"
 	"github.com/projectcalico/node/pkg/health"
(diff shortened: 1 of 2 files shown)


lmm (approved):
LGTM

PR #1254, opened 2021-10-14
Add CalicoUpgrade service for Calico upgrade on Windows

lmm:
## Description

This PR adds a new flag to calico-node that runs the new calico upgrade service for upgrades on Windows.
This PR is a continuation of https://github.com/projectcalico/node/pull/1204

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Add calico-node upgrade service for upgrades on Windows
```

### DIFF
--- a/pkg/upgrade/upgrade.go
+++ b/pkg/upgrade/upgrade.go
@@ -266,6 +266,16 @@ func verifyImagesSharePathPrefix(first, second string) error {
 		return nil
 	}
 
+	// Validation should catch this but just in case.
+	if len(n1PathParts) == 0 || len(n2PathParts) == 0 {
+		return fmt.Errorf("images %q and/or %q do not contain a valid path", first, second)
+	}
+
+	// If image paths do not have equal # of segments then they don't match.
+	if len(n1PathParts) != len(n2PathParts) {
+		return fmt.Errorf("images %q and %q do not share the same path prefix", first, second)
+	}
+
 	// Remove the last segment of the image path since it will contain the
 	// component name.
 	n1PathPrefix := n1PathParts[:len(n1PathParts)-1]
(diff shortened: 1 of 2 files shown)


pkg/upgrade/upgrade.go
```diff
@@ -0,0 +1,350 @@
...
+
+	// Split the image path. E.g. if the image is
+	// "docker.io/calico/node:v3.21.0" then the path is "calico/node".
+	n1PathParts := strings.Split(image.Path(n1), "/")
+	n2PathParts := strings.Split(image.Path(n2), "/")
+
+	// Remove the last segment of the image path since it will contain the
+	// component name.
+	n1PathPrefix := n1PathParts[:len(n1PathParts)-1]
+	n2PathPrefix := n2PathParts[:len(n2PathParts)-1]
```
song-jiang, line 264: I think it is possible the image name is `domain/calico-node:xxx` where n1PathParts has zero length.

pkg/upgrade/upgrade.go
```diff
@@ -0,0 +1,350 @@
...
+	n1, err := image.ParseNamed(first)
+	if err != nil {
+		return err
+	}
+	n2, err := image.ParseNamed(second)
+	if err != nil {
+		return err
+	}
+	// Compare the domain parts (e.g. docker.io) of the image references.
+	if image.Domain(n1) != image.Domain(n2) {
```
song-jiang, line 252: Are we sure if image.Domain(n1) is not ""?

pkg/upgrade/upgrade.go
lmm: Thanks, you're right. I fixed this to skip checking the path if the image reference is a short one (in this case the path must be the component name)

pkg/upgrade/upgrade.go
lmm: Yes, Domain will always be defined, reference.ParseNamed fails without a valid domain in the image reference.

pkg/upgrade/upgrade.go
song-jiang: The compare of short path is fine, but it is always good practice to add check for zero length on both image paths, that will make sure [:len(n1PathParts)-1] not panic. Also we don't need to proceed if the length is different? If n1PathParts has more elements than n2PathParts, it will panic to compare them one by one later. 
```
if (len(n1PathParts) == 0) || (len(n2PathParts) == 0) || (len(n1PathParts) != len(n2PathParts) ) {
  failed
}

song-jiang (approved):
LGTM

PR #1318, opened 2021-11-02
windows: verify upgrade image using node pod

lmm:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/winupgrade/upgrade.go
+++ b/pkg/winupgrade/upgrade.go
@@ -296,9 +296,7 @@ func verifyImagesSharePathPrefix(first, second string) error {
 
 func verifyPodImageWithHostPathVolume(cs kubernetes.Interface, nodeName string, hostPath string) error {
 	// Get pod list for calico-system pods on this node.
-	list, err := cs.CoreV1().Pods("calico-system").List(context.TODO(), metav1.ListOptions{
-		FieldSelector: "spec.nodeName=" + nodeName,
-	})
+	list, err := cs.CoreV1().Pods("calico-system").List(context.TODO(), metav1.ListOptions{})
 	if err != nil {
 		return err
 	}
@@ -310,11 +308,12 @@ func verifyPodImageWithHostPathVolume(cs kubernetes.Interface, nodeName string, 
 		if strings.HasPrefix(pod.Name, "calico-node") && pod.Spec.ServiceAccountName == "calico-node" {
 			calicoPod = pod
 			found = true
+			break
 		}
 	}
 
 	if !found {
-		return fmt.Errorf("could not find calico-node pod")
+		return fmt.Errorf("could not find a calico-node pod")
 	}
 
 	// Ensure the pod is owned by the calico-node daemonset.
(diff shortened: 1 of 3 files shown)


pkg/winupgrade/upgrade.go
```diff
@@ -295,14 +295,52 @@ func verifyImagesSharePathPrefix(first, second string) error {
...
+	}
+	if !ownedByCalicoDs {
+		return fmt.Errorf("calico-node pod not owned by calico-node daemonset")
+	}
+
+	// Get the calico node nodeImage
+	var nodeImage string
+	for _, c := range calicoPod.Spec.Containers {
+		if c.Name == "calico-node" {
+			nodeImage = c.Image
```
song-jiang, line 335: We should break here?

pkg/winupgrade/upgrade.go
```diff
@@ -295,14 +295,52 @@ func verifyImagesSharePathPrefix(first, second string) error {
 }
 
 func verifyPodImageWithHostPathVolume(cs kubernetes.Interface, nodeName string, hostPath string) error {
-	// Get pod list for all pods on this node.
-	list, err := cs.CoreV1().Pods("").List(context.TODO(), metav1.ListOptions{
+	// Get pod list for calico-system pods on this node.
+	list, err := cs.CoreV1().Pods("calico-system").List(context.TODO(), metav1.ListOptions{
 		FieldSelector: "spec.nodeName=" + nodeName,
```
song-jiang, line 300: I don't think it is correct that we use a nodeName selector since we are on a Windows node which has no calico-node pod.

pkg/winupgrade/upgrade.go
```diff
@@ -295,14 +295,52 @@ func verifyImagesSharePathPrefix(first, second string) error {
...
 		return err
 	}
 
+	// Find the calico-node pod.
+	var calicoPod v1.Pod
+	var found bool
+	for _, pod := range list.Items {
+		if strings.HasPrefix(pod.Name, "calico-node") && pod.Spec.ServiceAccountName == "calico-node" {
+			calicoPod = pod
+			found = true
```
song-jiang, line 312: We should break as soon as we found one.

pkg/winupgrade/upgrade.go
```diff
@@ -316,14 +354,6 @@ func verifyPodImageWithHostPathVolume(cs kubernetes.Interface, nodeName string,
...
-	nodeDs := daemonset("calico-node")
-	nodeImage, err := nodeDs.getContainerImage(cs, "calico-system", "calico-node")
-	if err != nil {
-		return err
-	}
-	log.Infof("Found node container image is %v\n", nodeImage)
-
 	// Walk through pods
 	for _, pod := range list.Items {
 		if !hasHostPathVolume(pod, hostPath) {
```
song-jiang, line 359: I'm thinking we should make sure only one pod mounted the host path. So I suggest we get that pod first.
```
var podWithHostPath *v1.pod
count := 0
for _, pod := range list.Items {
    if hasHostPathVolume(pod, hostPath) {
       podWithHostPath = pod
       count ++
    }
}

if count == 0 {
    return fmt.Errorf("Failed to find calico-windows-upgrade pod")
}
if count > 1 {
  return fmt.Errorf("More than one pod have host path")
}

// Continue to check image
...

```

song-jiang (changes requested):
A few comments.

song-jiang (approved):
LGTM

PR #1322, opened 2021-11-02
Update confd pins for disableBGPExport support

coutinhop:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


caseydavenport (approved):
Looks good, ideally we'd add a system test to this suite somewhere to validate the new function is actually present in the produced node images. That can come as a separate PR to expedite this though!

coutinhop:
> Looks good, ideally we'd add a system test to this suite somewhere to validate the new function is actually present in the produced node images. That can come as a separate PR to expedite this though!

Thanks, @caseydavenport! Yes, I'll add a test for it in a separate PR!

PR #1332, opened 2021-11-08
Increase node patch timeout for shutdown to 30 seconds.

song-jiang:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Increase timeout for setting node network unavailable on shutdown
```

### DIFF
--- a/pkg/lifecycle/shutdown/shutdown.go
+++ b/pkg/lifecycle/shutdown/shutdown.go
@@ -68,10 +68,10 @@ func Run() {
 				k8sNodeName = nodeRef
 			}
 
-			// Set node condition, the timeout value is 5 seconds.
+			// Set node condition, the timeout value is 30 seconds.
 			// Depends on how we configure termination grace period, this operation
 			// may not be successful if it takes too long to update node condition.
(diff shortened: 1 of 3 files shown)


caseydavenport (approved):

caseydavenport:
@song-jiang thinking about this more, I wonder if we should extend this quite a bit more. I'm thinking we can have a 5-10s timeout per-connection, but retry indefinitely even.

This is because the code in question is really only executed on teardown, which has a grace period and an eventual hard stop.

If we make our retry indefinite, we'll support any grace period duration (there's no harm in retrying more AFAIK). To speed things up, users can simply adjust the grace period to a shorter duration which will make the hard shutdown come earlier. WDYT?

song-jiang:
@caseydavenport I actually don't see too much benefit of extending it to more than 30 seconds. The worst case for shutdown process not working properly is that new pods are scheduled before calico-node setup the networking for them. However, I would argue that if calico-node failed to update node status for more than 30 seconds (something wrong with host <-> apiserver networking),  it is unlikely kubelet will get a new pod or calico-node will get complete policy sync.

PR #1337, opened 2021-11-10
IPv6: Increase blockSize maximum to 64 bits

olljanat:
## Description
Current IPv6 blockSize limit to 116 bits which was introduced by #339 is against of best practices:
> The IPv6 subnet size is standardized by fixing the size of the host identifier portion of an address to 64 bits.

- source: https://en.wikipedia.org/wiki/IPv6

Also articles like https://itnext.io/how-to-run-ipv6-enabled-docker-containers-on-aws-87e090ab0397 and https://itnext.io/kubernetes-multi-cluster-networking-made-simple-c8f26827813 highly recommend to use at least /80

Also there is known bugs on Docker which comes with those smaller networks like: https://github.com/moby/moby/issues/40275

So this PR will give possibility for user to configure bits between 64-115 too. Later it might make sense to change even default example to that /80 but I did leave it out for now.

Docs PR: https://github.com/projectcalico/calico/pull/5080

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/lifecycle/startup/startup.go
+++ b/pkg/lifecycle/startup/startup.go
@@ -603,14 +603,14 @@ func parseBlockSizeEnvironment(envValue string) int {
 
 // validateBlockSize check if blockSize is valid
 func validateBlockSize(version int, blockSize int) {
-	// 20 to 32 (inclusive) for IPv4 and 116 to 128 (inclusive) for IPv6
+	// 20 to 32 (inclusive) for IPv4 and 64 to 128 (inclusive) for IPv6
 	if version == 4 {
 		if blockSize < 20 || blockSize > 32 {
 			log.Errorf("Invalid blocksize %d for version %d", blockSize, version)
 			utils.Terminate()
 		}
 	} else if version == 6 {
-		if blockSize < 116 || blockSize > 128 {
+		if blockSize < 64 || blockSize > 128 {
 			log.Errorf("Invalid blocksize %d for version %d", blockSize, version)
 			utils.Terminate()
 		}


caseydavenport:
Thanks for the PR!

If I recall correctly, a smaller block size was chosen for Calico for a few reasons:
- 4096 is way more container IPs than is generally run on a node today.
- Calico supports allocating multiple blocks to a node - if you need more IP space, Calico will assign new blocks to the node on-demand - you're not limited by the block size, it's just the smallest increment that Calico will assign to a node.

> The IPv6 subnet size is standardized by fixing the size of the host identifier portion of an address to 64 bits.

In addition, Calico is a routed network, each pod gets a point-to-point link with the host. Don't think of the "host" as owning a subnet - all traffic from pod to any destination is routed over L3 (including its own host, and pods on the same host). In fact, Calico's routing model would work even if all of the container IPs on the node were disjoint and didn't not belong to a contiguous block.

For that same reason, the /80 limitation listed in the articles you linked does not apply to Calico. The way Calico configures the network is fundamentally different to how Docker configures networking - there's no L2 bridge connecting containers on the same host, and there's no NDP traffic between containers on the same host. proxy_ndp is set to ensure all IP traffic is forwarded to the host.

olljanat:
> Thanks for the PR!

Thanks for good comments. I'm new on IPv6 world so still learning what would be best. However some more questions/comments comes to my mind.

> If I recall correctly, a smaller block size was chosen for Calico for a few reasons:
> 
>     * 4096 is way more container IPs than is generally run on a node today.
> 
>     * Calico supports allocating multiple blocks to a node - if you need more IP space, Calico will assign new blocks to the node on-demand - you're not limited by the block size, it's just the smallest increment that Calico will assign to a node.

Both comments are valid but:
* One of the biggest reasons to go with IPv6 is make sure that environment will never run out of IPs. And when I said never I mean at least during my lifetime.
* Sure multiple blocks are possible but it is still kind of against of IPv6 idea where you never should run out of IPs.

> In addition, Calico is a routed network, each pod gets a point-to-point link with the host. Don't think of the "host" as owning a subnet - all traffic from pod to any destination is routed over L3 (including its own host, and pods on the same host).

I actually want let each node to own subnet because that is way how it is done on Kubernetes standard IPAM so my setup stays compatible with that or any other CNI/IPAM plugin which we might want use in future (I have hard time for choosing between Calico and Cilium currently).

Kubernetes default IPAM looks to be defaulting to /64 subnet per node https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/cmd/kube-controller-manager/app/options/nodeipamcontroller.go#L39-L41

> In fact, Calico's routing model would work even if all of the container IPs on the node were disjoint and didn't belong to a contiguous block.

That would end up to very big and complex routing tables on network infra which is something I want to avoid and which why I'm planning to use topology based IP pools.

> For that same reason, the /80 limitation listed in the articles you linked does not apply to Calico. The way Calico configures the network is fundamentally different to how Docker configures networking - there's no L2 bridge connecting containers on the same host, and there's no NDP traffic between containers on the same host. proxy_ndp is set to ensure all IP traffic is forwarded to the host.
> 
> Please let me know if that makes sense - I believe that means we don't need to extend the default block size, as we already support a flexible model that will allocate additional blocks on demand, and we're not impacted by the limitations of docker bridge networking described above.

What you are saying make sense on environment where Calico is only CNI. However I'm trying to create future proof, multi-cluster setup (kind of than on https://itnext.io/kubernetes-multi-cluster-networking-made-simple-c8f26827813 ) where different clusters can even have different CNI in use and if/when needed we can easily migrate workload from Kubernetes cluster to another which have totally different configuration. To way how I'm planning to achieve that is use publicly routable IPv6 addresses only for all pods and services together with service discovery using public DNS names.

caseydavenport:
> One of the biggest reasons to go with IPv6 is make sure that environment will never run out of IPs. And when I said never I mean at least during my lifetime.

Correct, and this is handled in Calico IPAM using multiple blocks. 

> Sure multiple blocks are possible but it is still kind of against of IPv6 idea where you never should run out of IPs.

I don't think it is - Calico doesn't run out of IPs, it allocates more on demand. 

> I actually want let each node to own subnet because that is way how it is done on Kubernetes standard IPAM

This isn't how Calico IPAM works - you can use host-local IPAM with Calico if you'd like. There's a known issue with it for IPv6 though I believe: https://github.com/projectcalico/cni-plugin/issues/1177. Should be an easy improvement though. 

> That would end up to very big and complex routing tables on network infra which is something I want to avoid and which why I'm planning to use topology based IP pools.

You'll end up with one route per block of 4096 addresses - that's not very many. In practice, that's one route per node, no more. Calico does route aggregation in its advertisements, but at the network layer there is no "subnettting" or L2 network connecting all the pods.

It sounds like you're looking for Calico IPAM to behave like something other than Calico IPAM, which is fine but I think is out of scope. If you're looking for host-local type allocations, Calico is compatible with other IPAM plugins, and that's the approach I'd recommend going.

PR #1338, opened 2021-11-10
Improve handling CalicoUpgrade service

song-jiang:
## Description

- Fix log spam for Calico upgrade service.
- Allow uninstalling upgrade service manually.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/winupgrade/should_upgrade.go
+++ b/pkg/winupgrade/should_upgrade.go
@@ -16,6 +16,9 @@ package winupgrade
 import (
 	"context"
 	"os"
+	"strings"
+
+	"github.com/projectcalico/libcalico-go/lib/names"
 
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
@@ -31,7 +34,7 @@ func ShouldInstallUpgradeService() {
 	variant := getVariant()
 
 	// Determine the name for this node.
-	nodeName := utils.DetermineNodeName()
+	nodeName := determineNodeName()
 	log.Debugf("Check if Calico upgrade service should be installed on node: %s. Version: %s, Variant: %s, baseDir: %s", nodeName, version, variant, baseDir())
 
 	config, err := clientcmd.BuildConfigFromFlags("", kubeConfigFile())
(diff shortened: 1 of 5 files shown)


pkg/winupgrade/should_upgrade.go
```diff
@@ -25,13 +25,34 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+func GetLoggingLevel() log.Level {
+	rawLogLevel := os.Getenv("CALICO_STARTUP_LOGLEVEL")
```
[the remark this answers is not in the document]
caseydavenport: Is STARTUP_LOGLEVEL the right option for this? This isn't in the startup package.

pkg/winupgrade/should_upgrade.go
```diff
@@ -25,13 +25,34 @@ import (
...
+			log.WithError(err).Error("Failed to parse log level, defaulting to info.")
+		}
+	}
+
+	return log.InfoLevel
+}
+
 // Exit with code zero if Windows upgrade service should be installed.
 func ShouldInstallUpgradeService() {
+	// Don't log anything with INFO level because
```
caseydavenport, line 44: This isn't how we should do this, I don't think. It's not how we run any of the rest of our code. If the user specified INFO, we should log INFO level.

Any repetitive / spammy logs should be DEBUG level. Info should not be used for any repetitive logging.

song-jiang: The problem here is `utils.DetermineNodeName` calls `log.Info`. I would like to avoid copying over the entire function but at the same time to avoid adding `ShouldLog` parameters to the function.Hmm...

caseydavenport: Just downgrade that to debug. If we really need an info log we can move it out of that function somewhere else. 

Or, can you just call determine nodename once in `init` and use it throughout rather than calling on repeat?

song-jiang: > can you just call determine nodename once in init and use it throughout rather than calling on repeat?
Good idea.

song-jiang: Oh...Wait. The `calico-node -shoud-upgrade` is called every 10 seconds so I suppose `init()` function will be called every 10 seconds as well.

caseydavenport: well, the idea still stands - we can call determineNodename outside of the loop and pass it in as an argument / cli flag / whatever. I don't have the full context about how this code is actually called.

Alternatively, just move the log line out of the `DetermineNodename`. It's probably a bad idea to have INFO level logging in a helper function anyway

song-jiang: It is still very useful to be able to print node name and how it is determined. I've updated the function to return the source for the node name and replaced `log.Info` to`log.Debug`

caseydavenport: I'm not a huge fan of modifying the function to return a source just for logging purposes. Ideally, we'd just call `DetermineNodename` outside of whatever cyclical thing is happening so that we get logging at the right place and at the right level. It's also not really worth jumping through hoops for - it's fine at debug level. In 99% of cases it's not needed, except if you're debugging the cluster, in which case you can enable debug logging, right? 

For my own knowledge, what is calling `calico-node -should-upgrade` called every 10 seconds?

caseydavenport: Actually, I found it here: https://github.com/projectcalico/node/blob/b5f26bcdb221512f3c7d1a2639e2bdca4c314536/windows-packaging/CalicoWindows/node/node-service.ps1#L176-L180

song-jiang: From my experience, nodename is quite essential for Calico. The first thing to debug a cluster, especially on Windows node, is to see if user got the right nodename.  I really like to keep the way it is now that we can clearly see how nodename is set at info level, except for upgrade service. 

I don't like the code changes either and seems to be overcomplicated. Can I propose of copying `util.DetermineNodename` function to `winupgrade` package with some comments? 
- Nothing changes on other files. 
- Upgrade service is a temporary workaround for Windows, it will get deprecated anyway with daemonset install later. 
WDYT?

lmm (approved):

PR #1371, opened 2021-11-30
windows: pass ExceptUpgradeService param to stop-calico.ps1 as well

lmm:
## Description

When stop-calico.ps1 is called from uninstall.ps1 we have to also pass
in the ExceptUpgradeService param to stop-calico.ps1
Otherwise the calico-upgrade service is stopped.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


song-jiang (approved):
LGTM. Thanks for fixing this

